PR #4, opened 2020-09-10
run tests, deploy to test server, install and run tests

thigm85:
Every push will trigger unit tests and integration tests. When merged to master, the package will be published to test PyPI and then we install the package from the test server and run the tests again to ensure the library works as expected.


setup.py
```diff
@@ -3,9 +3,9 @@
 
 
 def get_target_version():
-    build_nr = os.environ.get("GITHUB_RUN_NUMBER", "0+dev")
-    version = "0.1"
-    return "{}.{}".format(version, build_nr)
+    pyvespa_version = os.environ.get("PYVESPA_VERSION", "0+dev")
+    build_nr = os.environ.get("TRAVIS_BUILD_NUMBER", "0+dev")
+    return "{}.{}".format(pyvespa_version, build_nr)
```
oyving, line 8: I don't think `0+dev.0+dev` is a valid version format, so I think you need a different default for `pyvespa_version`.

oyving (changes requested):
There are some inconsistent indent levels in your lists in the YAML file which would be good to clean up as well.

oyving (approved):

PR #8, opened 2020-09-16
make it possible to submit key content instead of key file

thigm85:

### DIFF
--- a/vespa/package.py
+++ b/vespa/package.py
@@ -540,17 +540,18 @@ class VespaCloud(object):
     def _read_private_key(
         key_location: Optional[str] = None, key_content: Optional[str] = None
     ) -> ec.EllipticCurvePrivateKey:
-        if not key_content:
+
+        if key_content:
+            key_content = bytes(key_content, "ascii")
+        elif key_location:
             with open(key_location, "rb") as key_data:
                 key_content = key_data.read()
         else:
-            key_content = bytes(key_content, "ascii")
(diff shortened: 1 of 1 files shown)


vespa/package.py
```diff
@@ -534,16 +537,21 @@ def __init__(
...
-            key = serialization.load_pem_private_key(
-                key_data.read(), None, default_backend()
+    def _read_private_key(
+        key_location: Optional[str] = None, key_content: Optional[str] = None
+    ) -> ec.EllipticCurvePrivateKey:
+        if not key_content:
+            with open(key_location, "rb") as key_data:
+                key_content = key_data.read()
+        else:
+            key_content = bytes(key_content, "ascii")
```
oyving, line 547: You should check both `key_content` and `key_location` and fail with an error message if both are `None`.

vespa/package.py
```diff
@@ -534,16 +537,21 @@ def __init__(
...
+
+        key = serialization.load_pem_private_key(key_content, None, default_backend())
+        if not isinstance(key, ec.EllipticCurvePrivateKey):
+            raise TypeError(
+                "Key at " + key_location + " must be an elliptic curve private key"
```
oyving, line 552: Update error message - it might be that `key_content` is the wrong type of key.

oyving (changes requested):

oyving (approved):

PR #10, opened 2020-09-28
Normalized Discounted Cumulative Gain evaluation metric

thigm85:
* Implement nDCG metric (reference: https://en.wikipedia.org/wiki/Evaluation_measures_(information_retrieval)#Discounted_cumulative_gain)
* Add unit test
* I also added one more example in the unit tests with score > 1 to make sure all the metrics do sensible things in this case.


lesters (approved):
Great!

PR #25, opened 2020-10-14
track data collection progress

thigm85:
This PR enable some information when collecting training data from Vespa apps. Before this PR the following code

```
training_data_batch = app.collect_training_data(
    labelled_data = train_set,
    id_field = "cord_uid",
    query_model = query_models["or_bm25"],
    number_additional_docs = 9,
    fields = ["title-full"]
)
```
would processed many queries and documents without indicating how far into the process the call was. So the user did not know if the process was still running.

This PR adds the possibility of printing progress statement of the following form at every `show_progress` iteration.

`Query {}/{}, Doc {}/{}. Query id: {}. Doc id: {}`

### DIFF
--- a/vespa/application.py
+++ b/vespa/application.py
@@ -1,6 +1,7 @@
 # Copyright Verizon Media. Licensed under the terms of the Apache 2.0 license. See LICENSE in the project root.
 
-from typing import Optional, Dict, Tuple, List
+import sys
+from typing import Optional, Dict, Tuple, List, IO
 from pandas import DataFrame
 from requests import post
 from requests.models import Response
@@ -16,6 +17,7 @@ class Vespa(object):
         port: Optional[int] = None,
         deployment_message: Optional[List[str]] = None,
         cert: Optional[str] = None,
+        output_file: IO = sys.stdout,
     ) -> None:
         """
         Establish a connection with a Vespa application.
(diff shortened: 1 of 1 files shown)


vespa/application.py
```diff
@@ -213,14 +218,36 @@ def collect_training_data(
...
         :param kwargs: Extra keyword arguments to be included in the Vespa Query.
         :return: DataFrame containing document id (document_id), query id (query_id), scores (relevant)
             and vespa rank features returned by the Query model RankProfile used.
         """
 
         training_data = []
+        number_queries = len(labelled_data)
+        idx_total = 0
+        idx_query = 0
         for query_data in labelled_data:
```
oyving, line 232: You can do the counting (both for queries and docs) by using a count iterator instead of keeping track your self.
```python
    for query_idx, query_data in zip(itertools.count(), labelled_data):
        ...
```

oyving (approved):
Just added a way to use iterators to remove some counter tracking, but otherwise LGTM.

vespa/application.py
oyving: And I see the even easier way is to use `enumerate`.
```python
   for query_idx, query_data in enumerate(labelled_data):
      ...
```

thigm85: True. Thanks.

PR #39, opened 2020-11-03
Allow users to specify custom Docker port

thigm85:

### DIFF
--- a/vespa/package.py
+++ b/vespa/package.py
@@ -389,6 +389,7 @@ class ApplicationPackage(ToJson, FromJson["ApplicationPackage"]):
 class VespaDocker(object):
     def __init__(
         self,
+        port: int = 8080,
         output_file: IO = sys.stdout,
     ) -> None:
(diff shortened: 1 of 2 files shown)


vespa/package.py
```diff
@@ -418,7 +419,7 @@ def _run_vespa_engine_container(
                     hostname=application_name,
                     privileged=True,
                     volumes={disk_folder: {"bind": "/app", "mode": "rw"}},
-                    ports={self.local_port: self.local_port, 19112: 19112},
+                    ports={8080: self.local_port},
```
oyving, line 422: So this is saying map container port 8080 to `self.local_port` right? If so, that looks ok.

oyving (approved):

vespa/package.py
thigm85: Yes. This was non-intuitive for me. I would assume the opposite direction but got here through trial and error.

PR #52, opened 2020-11-11
Add default query profile with root query profile type by default to all apps

thigm85:
Including default query profile and root query profile type by default so that the user does not have to specify it every time he creates an application. A future PR will allow them to include fields on the query profile and query profile type via pyvespa API.


thigm85:
update

oyving (approved):

PR #68, opened 2021-01-30
Add a BERT model config class that will later be used to include BERT into ranking

thigm85:


thigm85:
@lesters Please review. I left one failing test that I want to discuss with you because the test is important and your ONNX experience might help. 

In the test, I export a sample BERT model to ONNX. I then compare the prediction made by loading the ONNX model and running it with ONNX runtime with the prediction made directly by the model that I exported. Those two predictions should be the same (up to rounding error) but they are not and I am not sure why.

thigm85:
@oyving please review.

oyving (approved):

setup.py
```diff
@@ -23,7 +23,7 @@ def get_target_version():
     ),
     packages=setuptools.find_packages(),
     include_package_data=True,
-    install_requires=["requests", "pandas", "docker", "jinja2", "cryptography"],
+    install_requires=["requests", "pandas", "docker", "jinja2", "cryptography", "torch", "transformers", "onnxruntime"],
```
thigm85, line 26: @lesters @oyving 

I feel it is overkill to add those three libraries as dependencies to pyvespa. 

* `torch` and `transformers` are required if the user wants to build applications with BERT model.
* `onnxruntime` is only used for a unit test.

Do you think that implementing something like [this](https://github.com/huggingface/transformers/blob/master/src/transformers/convert_graph_to_onnx.py#L267-L271) is more sensible than requiring everyone to install all those libraries?

lesters (approved):
LGTM. With the change to using `convert_pytorch` I expect the test to work as intended.

setup.py
lesters: I think that is reasonable, but these are requirements for passing the suit of unit tests run in the build process?

thigm85: Yes, but I can have a requirements.txt file responsible for installing the optional packages during the CI pipeline.

oyving: Yeah, I think that makes sense. Make the dependencies marked as extras in the file and use requirements.txt to make sure everything is in place during testing.

You should also make sure we have tests that test the behavior if the dependencies are absent.

oyving: I guess you are going to follow this guide? https://setuptools.readthedocs.io/en/latest/userguide/dependency_management.html#optional-dependencies

lesters: That does sound sensible. +1

thigm85:
@lesters @oyving Can you take a look at the last two commits. It implements the optional dependency discussed above.

lesters:
LGTM!

PR #104, opened 2021-03-09
Add save and load method for ApplicationPackage

thigm85:

### DIFF
--- a/vespa/json_serialization.py
+++ b/vespa/json_serialization.py
@@ -1,6 +1,7 @@
 import datetime
 import json
 import typing
+from pkg_resources import get_distribution
 
 T = typing.TypeVar("T")
 
@@ -15,13 +16,13 @@ class ToJson(object):
     to understand what has happened.
 
         {
-          "version": 1,
+          "version": 0.1.0,
           "class": "Field",
           "serialized_at": "2018-10-24T12:55:32+00:00",
           "data": { ... }
         }
 
-        * version: This value is hard-coded to 1.
+        * version: pyvespa package version.
         * class: The name of the class we serialized.  For debugging purposes.
         * serialized_at: The time we serialized the instance of the class.  For debugging purposes.
         * data: The actual data of the serialized class.
@@ -37,7 +38,7 @@ class ToJson(object):
     @property
     def to_envelope(self) -> typing.Mapping:
         return {
-            "version": 1,
+            "version": get_distribution("pyvespa").version,
             "class": self.__class__.__name__,
             "serialized_at": datetime.datetime.utcnow().isoformat(),
             "data": self.to_dict,
--- a/setup.py
+++ b/setup.py
@@ -3,7 +3,7 @@ import setuptools
 
 
 def get_target_version():
-    pyvespa_version = os.environ.get("PYVESPA_VERSION", "0.2")
+    pyvespa_version = os.environ.get("PYVESPA_VERSION", "0.3")
     build_nr = os.environ.get("TRAVIS_BUILD_NUMBER", "0+dev")
     return "{}.{}".format(pyvespa_version, build_nr)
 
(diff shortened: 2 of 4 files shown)


vespa/json_serialization.py
```diff
@@ -37,7 +38,7 @@ def to_dict(self) -> typing.Mapping:
     @property
     def to_envelope(self) -> typing.Mapping:
         return {
-            "version": 1,
+            "version": get_distribution("pyvespa").version,
```
thigm85, line 41: @oyving Do you agree with this? I think it makes more sense to tag the serialized object with the pyvespa version used to create it instead of hardcode a version number.

oyving: The version number here should be the version number of the format and not the version number of pyvespa. This makes it easier to see what deserialisation code to invoke later should you wish to support multiple versions of the format.

oyving (commented):
I am not sure about the approach of serialising the application package to JSON when we already have an application package format. I understand what we have is not a convenient format for these experiments, but I think adding a new one is something that should be discussed with @bratseth so we have agreement on on-disk formats for applications.

thigm85:
I understand the point. We can export from pyvespa to the Vespa application package format. However, importing it back from the Vespa application package to pyvespa is more painful because pyvespa provides just a Vespa API subset. The save/load method described here also includes ml-related classes like model config classes that are not properly covered by the Vespa application package format. With that said, I agree this needs more discussion. @bratseth @oyving 

> I am not sure about the approach of serialising the application package to JSON when we already have an application package format. I understand what we have is not a convenient format for these experiments, but I think adding a new one is something that should be discussed with @bratseth so we have agreement on on-disk formats for applications.

vespa/json_serialization.py
thigm85: But what if the `to_dict` and `from_dict` methods for a given class change in pyvespa version 0.3.5 for example. How will I track which version was used by old serialized objects?

bratseth:
The application package is a user interface, I think this is just an internal serialization format the users will never see? What's used then is more about what's most convenient to implement.

That said, it could be simpler to use a regular application package and add some more files to it (generally legal), since you already have (lossy) application package import/export.

vespa/json_serialization.py
oyving: I guess it depends on the level of compatibility and strictness you want to follow. If you always make sure later versions understand earlier versions, then just comparing the version number of pyvespa will be fine. You might have to track version ranges inside `from_dict` if the format changes in some significant way in the future.

thigm85: Yes. Making sure later versions understand earlier versions is nice. But at the very least it would be nice to track which pyvespa version was used to serialize the object so that people can go back to earlier versions if needed. That is why I included `get_distribution("pyvespa").version`. 

Maybe have `library_version` in addition to the (serialization format) `version` and add it on the `from_dict` signature?
E.g.
```
mapping_class.from_dict(mapping["data"], mapping["library_version"])
```

thigm85:
> The application package is a user interface, I think this is just an internal serialization format the users will never see? What's used then is more about what's most convenient to implement.
> 
> That said, it could be simpler to use a regular application package and add some more files to it (generally legal), since you already have (lossy) application package import/export.

Yes, this PR is about an internal serialization format. The current version is much more convenient to implement as what needs to be imported is limited by what can be exported.

I agree that importing a regular application package into pyvespa would be ideal as the internal format would coincide with the familiar user interface already used. However, this needs to be explored as it is difficult to know how hard it would be to implement such an import method. My initial guess is that it is substantially harder and error-prone.

vespa/json_serialization.py
oyving: I think both are acceptable solutions. If we're not currently using the version number in the format for anything it would be OK to change it to be the library version and then handle those in code later should we need to.

oyving (approved):

PR #112, opened 2021-03-19
Lesters/async feeding 

lesters:
@thigm85 Please take a look. For discussion.

Conceptually, you call `feed_batch`, which takes care of an asyncio loop. If the user already has such a loop, they can call the `async_feed_batch` directly, or even `async_feed_data_point` if they already have a `aiohttp` session themselves.

We should discuss API for this, and by extension, for querying as well.


vespa/application.py
```diff
@@ -144,6 +148,55 @@ def feed_data_point(self, schema: str, data_id: str, fields: Dict) -> Response:
...
+        :param fields: Dict containing all the fields required by the `schema`.
+        :return: Response of the HTTP POST request.
+        """
+        end_point = "{}/document/v1/{}/{}/docid/{}".format(
+            self.end_point, schema, schema, str(data_id)
+        )
+        vespa_format = {"fields": fields}
+        sslcontext = False
+        if self.cert is not None:
+            sslcontext = ssl.create_default_context(cafile=self.cert)
```
thigm85, line 167: We should add a cloud integration test to check if the `sslcontext` works as expected.

thigm85 (dismissed):

thigm85:
@lesters Let me know if you can run the cloud test locally now that you have the key.

vespa/application.py
lesters: Fixed.

lesters:
@thigm85 Was able to run cloud test locally and test the cloud batch feeding.

vespa/test_integration_docker.py
```diff
@@ -236,6 +237,129 @@ def test_data_operation(self):
...
+            body = f"body for document {i}"
+            fields = {"id": id, "title": title, "body": body}
+            docs.append((schema, id, fields))
+        app.feed_batch(docs)
+
+        # Verify that all documents are fed
+        result = app.query(query="sddocname:msmarco", query_model=QueryModel())
+        self.assertEqual(result.number_documents_indexed, num_docs)
+
+    def test_data_operations_async(self):
```
thigm85, line 262: @lesters I get the following warning when I run the async tests:

```
/Users/tmartins/.local/share/virtualenvs/pyvespa-UOfd_sG0/lib/python3.7/site-packages/urllib3/util/url.py:421: ResourceWarning: unclosed <socket.socket fd=15, family=AddressFamily.AF_UNIX, type=SocketKind.SOCK_STREAM, proto=0, raddr=docker.sock>
  fragment=ensure_type(fragment),
``` 
Could it be that the `self.aiohttp_session.close()` is not closing all the used connections?

PR #118, opened 2021-04-26
Make QueryModel more flexible by adding body_function argument

thigm85:
The `body_function` argument allows us to specify the body of the request using the Vespa Query API

```
def body_function(query):
    body = {'yql': 'select * from sources * where userQuery();',
            'query': query,
            'type': 'any',
            'ranking': {'profile': 'bm25', 'listFeatures': 'true'}}
    return body

query_model = QueryModel(body_function = body_function)

res = app.query(query = "this is a test", query_model=query_model)
```

### DIFF
--- a/vespa/query.py
+++ b/vespa/query.py
@@ -129,7 +129,7 @@ class ANN(MatchFilter):
         query_vector: str,
         hits: int,
         label: str,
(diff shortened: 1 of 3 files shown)


lesters (approved):
LGTM!

PR #119, opened 2021-04-27
Make it possible to format query results

thigm85:
Allow formatting query results according to TREC format.

Skjermbilde 2021-04-27 kl  09 19 52


lesters (approved):
LGTM!

PR #120, opened 2021-04-28
support df as input to app.evaluate

thigm85:
Before this PR we only accepted a list of Dict as input to app.evaluate:

Skjermbilde 2021-04-28 kl  06 24 27

This PR allows the user to use a more familiar df format as input:

Skjermbilde 2021-04-28 kl  06 26 33


lesters (approved):
LGTM!

PR #121, opened 2021-04-28
Simplified metrics output by default with option for detailed metrics

thigm85:
Evaluation returns just the basic metric values by default. The option `detailed_metrics` allow intermediate metric values to be returned as well.

Skjermbilde 2021-04-28 kl  08 01 40


lesters (approved):
👍

PR #122, opened 2021-04-28
Evaluate multiple query models

thigm85:
We can now give a name to QueryModel. This will be used to identify the model when computing metrics.

Skjermbilde 2021-04-28 kl  20 54 41

We can evaluate a single model as before.

Skjermbilde 2021-04-28 kl  21 00 43

Or compare multiple models:

Skjermbilde 2021-04-28 kl  21 06 56


lesters (approved):
👍

PR #124, opened 2021-04-29
Enable per_query argument. Default to per model summary.

thigm85:
We return summary statistics per model by default:

Skjermbilde 2021-04-29 kl  08 59 08

We can choose which summary statistics to use:

Skjermbilde 2021-04-29 kl  09 00 28

Or return raw metrics per (model, query_id):

Skjermbilde 2021-04-29 kl  09 02 06


lesters (approved):
👍

PR #138, opened 2021-05-12
Easier sync/async API for batch jobs

thigm85:
I want to propose a simpler API for async/sync batch jobs. I exemplify the API for feeding in this PR.

Calling `app.feed_batch` run the job synchronously by default:

Skjermbilde 2021-05-12 kl  11 27 00

Adding `asynchronous=True` will run the job asynchronously **IF** there is no running loop in place (see for example [this integration test](https://github.com/vespa-engine/pyvespa/pull/138/files#diff-e975894cbf9c72564470cfc48307f5a2a657e531303a628df8ea1996b8942669R262-R278))

Skjermbilde 2021-05-12 kl  11 28 52

Adding `asynchronous=True` will return the coroutine in case there is a running loop in place (as is the case in jupyter notebooks). In this case, we can run the coroutine by using the `await` keyword:

Skjermbilde 2021-05-12 kl  11 33 01


vespa/application.py
```diff
@@ -210,14 +214,35 @@ def feed_data_point(self, schema: str, data_id: str, fields: Dict):
...
+        return [
+            self.feed_data_point(schema, data_point["id"], data_point["fields"])
+            for data_point in batch
+        ]
+
+    async def _feed_batch_async(self, schema: str, batch: List[Dict]):
+        async with VespaAsync(self) as async_app:
+            return await async_app.feed_batch(schema=schema, batch=batch)
+
+    def feed_batch(self, schema: str, batch: List[Dict], asynchronous=False):
```
lesters, line 227: I think one usually wants to feed as fast as possible, so unless there are negative consequences, the `asynchronous` arguments should default to `True`. Or better, `synchronous=False`.

lesters (approved):
LGTM. Consider comment about `asynchronous` to default `True` in batch feed.

vespa/application.py
thigm85: The only reason that I set `asynchronous=False` is that `asynchronous=True` requires an extra step from the user when using it where there is already a running loop in place, such as in the Jupyter notebooks. In those cases, the user needs to add an await keyword for the coroutine to run. What do you think?

PR #152, opened 2021-05-31
Fix bug in the NDCG calculation

thigm85:
Caught and fixed a bug on the NDCG calculation when comparing results between pyvespa and this [Vespa news tutorial](https://docs.vespa.ai/en/tutorials/news-5-recommendation.html).


lesters (approved):
👍

PR #166, opened 2021-06-12
Standardize sync and async output

thigm85:
Make sync and async operations return the same object. Required step to make running sync or async indifferent to the end-user.


lesters (approved):
Great!

PR #187, opened 2021-08-05
Implement get_batch with async mode as default

thigm85:
Now that we got `app.feed_batch` to work as expected, we will replicate the same behavior for other operations. This PR starts with `get_batch`.


lesters (approved):
LGTM!

PR #197, opened 2021-08-31
SequenceClassification task in Vespa

thigm85:
* Adapt `ApplicationPackage` to allow for simpler configurations, e.g. no schema. 
* Implement `ModelServer`, which is a simplified `ApplicationPackage` focused on stateless model evaluation.
* `ApplicationPackage` is now included on the `Vespa` instance when deploying apps with pyvespa.
  * This helps to process input/output involved when using stateless model evaluations through `app.predict`. 
* `Vespa` now has `get_model_endpoint` and `predict` to use for stateless model evaluation mode.
* Both `VespaDocker` and `VespaCloud` now work with `ModelServer` deployment to deploy stateless model evaluation apps.
* Created `SequenceClassification` task that inherits from the base class `TextTask`.

### DIFF
--- a/vespa/application.py
+++ b/vespa/application.py
@@ -18,6 +18,7 @@ from tenacity import retry, wait_exponential, stop_after_attempt
 from vespa.io import VespaQueryResponse, VespaResponse
 from vespa.query import QueryModel
 from vespa.evaluation import EvalMetric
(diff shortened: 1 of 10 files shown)


lesters (approved):
Brilliant!

PR #198, opened 2021-09-02
Improve naming of parameters associated with stateless model evaluation

thigm85:
* Use `tasks` instead of `models` as argument name for `ModelServer` to align with the fact this argument take instances of type `TextTask`.
* Use `model_id` instead of `model_name` to identify a model since `model_id` is what we use when defining a Task.


lesters (approved):
:+1:

PR #204, opened 2021-09-13
Remove ml module dependency from package module

thigm85:
This PR removes `vespa.package` module dependency on the `vespa.ml` module. The main goal is to avoid installing `vespa.ml` specific dependencies through `pip install pyvespa[ml]` unless the user explicitly wants to use `vespa.ml` classes.


lesters (approved):
LGTM!

PR #210, opened 2021-09-23
Infer schema name whenever possible for batch document operations

thigm85:
When we have application package information containing just one schema, it is possible to simplify the batch document operations by not specifying the schema name. For those cases, we can for example use `app.feed_batch(docs)` instead of `app.feed_batch(docs, schema = "sentence")` when we know there is only one schema named "sentence".


lesters (approved):
:+1:

PR #227, opened 2021-12-09
Use correct path for documentation

kkraune:
- tip: using linebreaks in doc makes later PRs easier to review ;-)


thigm85 (approved):
LGTM

PR #233, opened 2022-02-02
Feed to Documentation Search

kkraune:
... or @hmusum 

This is a copy of files from `documentation` with mods in the _plugins-vespafeed/vespa_index_generator.rb + config. A GitHub action will push on each commit


hmusum (approved):
Not familiar with feed setup and code, but LGTM

kkraune:
lets give it a go

PR #257, opened 2022-03-17
don't fail on input

kkraune:


thigm85:
Not sure the best way to handle this is to allow errors to happen

kkraune:
I agree. there might be alternatives. can also stop using user input. we will see if this fixes it, though

Issue #245, opened 2022-02-21
Make it possible to specify the container image to use in the VespaDocker class


aressem:
In issue https://github.com/vespa-engine/pyvespa/issues/231 it would have been nice to specify the container image to use. This is not only for custom built or previews, but some users might also want to use the GitHub Docker Registry instead of Docker Hub (`ghcr.io/vespa-engine/vespa`) for container registry.

thigm85:
@kkraune pyvespa 0.16.0 is out with the latest changes. 

Release instruction added on https://github.com/vespa-engine/pyvespa/pull/293, PTAL.

kkraune:
VespaDocker now supports a container_image argument: https://pyvespa.readthedocs.io/en/latest/reference-api.html#vespadocker

PR #292, opened 2022-04-01
Kkraune/container image

kkraune:
- ref https://github.com/vespa-engine/pyvespa/issues/245
- Add parameter to set custom Docker image + integration test

@thigm85 I am not sure if this is all we need to add, review carefully

### DIFF
--- a/vespa/test_integration_docker.py
+++ b/vespa/test_integration_docker.py
@@ -132,8 +132,8 @@ def create_sequence_classification_task():
 
 
 class TestDockerCommon(unittest.TestCase):
-    def deploy(self, application_package, disk_folder, container_image=""):
-        if len(container_image) > 0:
+    def deploy(self, application_package, disk_folder, container_image=None):
+        if container_image:
             self.vespa_docker = VespaDocker(port=8089, disk_folder=disk_folder, container_image=container_image)
         else:
             self.vespa_docker = VespaDocker(port=8089, disk_folder=disk_folder)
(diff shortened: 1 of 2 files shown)


vespa/test_integration_docker.py
```diff
@@ -132,8 +132,12 @@ def create_sequence_classification_task():
 
 
 class TestDockerCommon(unittest.TestCase):
-    def deploy(self, application_package, disk_folder):
-        self.vespa_docker = VespaDocker(port=8089, disk_folder=disk_folder)
+    def deploy(self, application_package, disk_folder, container_image=""):
```
thigm85, line 135: In python, it is standard practice to use `container_image = None` in this case.

vespa/test_integration_docker.py
```diff
@@ -132,8 +132,12 @@ def create_sequence_classification_task():
...
+        if len(container_image) > 0:
```
thigm85, line 136: When using `container_image = None` as per my suggestion above, all that is needed is to use `if container_image:`

vespa/test_integration_docker.py
```diff
@@ -1088,6 +1092,20 @@ def tearDown(self) -> None:
         self.vespa_docker.container.stop()
         self.vespa_docker.container.remove()
 
+class TestQaDockerDeploymentImage(TestDockerCommon):
+    def setUp(self) -> None:
+        self.app_package = create_qa_application_package()
+        self.disk_folder = os.path.join(os.getenv("WORK_DIR"), "sample_application")
+
+    def test_deploy(self):
```
thigm85, line 1100: Instead of creating a new class we can move this `test_deploy` method to the class `TestQaDockerDeployment` that already exists and call it `test_deploy_image`

thigm85 (changes requested):

vespa/test_integration_docker.py
kkraune: agreed, that will simplify

kkraune:
Great feedback! I also made the serialization skip the image tag, as we cannot really know today which tag is used in deploy

kkraune:
@thigm85 PTAL

kkraune:
all integration tests OK on my laptop now, PTAL @thigm85

PR #301, opened 2022-04-07
Kkraune/fix notebook testing ii

kkraune:
... or @jobergum 

I think these are the last ones. pipeline will now go red on failure


thigm85 (approved):
LGTM

PR #305, opened 2022-04-07
Wait for app load

kkraune:
- Sorry for the code duplication - VespaCloud and VespaDocker should inherit such functions from a base class. Let's refactor later, get the integration tests fixed first (now failed as we removed the retries in app.get_application_status)

### DIFF
--- a/vespa/deployment.py
+++ b/vespa/deployment.py
@@ -278,7 +278,7 @@ class VespaDocker(ToJson, FromJson["VespaDocker"]):
         try_interval = 5
         waited = 0
         while not self._check_configuration_server() and (waited < max_wait):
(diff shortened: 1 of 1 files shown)


thigm85 (approved):
I agree. We should refactor to reuse the code. This can be done when unifying Docker and Cloud deployment.

kkraune:
Thanks! I don't think secrets work in PR builds (so cloud integration tests will fail), so I will just merge this and see

thigm85:
FYI: Cloud integration tests currently do not run on PRs because we reuse the same Vespa Cloud instance. So running two PRs in parallel would lead to cloud integration tests errors.

PR #306, opened 2022-04-07
Refactor and fix the test

kkraune:


vespa/deployment.py
```diff
@@ -267,14 +267,19 @@ def _execute_deployment(
             deployment_message=deployment_message,
             application_package=application_package,
         )
-
-        self.wait_for_application_up(app, max_wait=300)
+        app.wait_for_application_up(max_wait=300)
```
thigm85, line 270: I see that we repeat the number 300 across the codebase. Could we have a global variable such as `DEPLOY_MAX_WAIT=300` defined at the top of the `deployment.py`?

thigm85 (changes requested):

vespa/deployment.py
kkraune: Sure. I was not sure whether to have one for image start + config server startup, and another for container start. The latter some times take longer. I guess we can have two variables. I'll fix that in next PR, but would like to have all pipelines green before doing more changes, makes development easier ...

thigm85 (approved):

PR #307, opened 2022-04-08
Kkraune/misc

kkraune:
- hopefully this fixes the could integration test
- will remove debug once fixed

### DIFF
--- a/vespa/application.py
+++ b/vespa/application.py
@@ -215,6 +215,8 @@ class Vespa(object):
         :return:
         """
         endpoint = "{}/ApplicationStatus".format(self.end_point)
+        if self.key:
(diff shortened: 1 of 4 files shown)


vespa/application.py
```diff
@@ -215,6 +215,8 @@ def get_application_status(self) -> Optional[Response]:
         :return:
         """
         endpoint = "{}/ApplicationStatus".format(self.end_point)
+        if self.key:
+            self.cert = (self.app.cert, self.app.key)
```
thigm85, line 219: This was handled by VespaSync before. I understand that you tried to simplify the get_application_status code in a previous PR but this might lead to duplication of code that is harder to maintain later.

thigm85 (approved):

PR #313, opened 2022-04-11
Add zip methods for deployment

kkraune:


vespa/package.py
```diff
@@ -1415,6 +1417,57 @@ def load(disk_folder: str) -> "ApplicationPackage":
         with open(file_path, "r") as f:
             return ApplicationPackage.from_json(f.read())
 
+    def to_zip(self) -> BytesIO:
```
thigm85, line 1420: My understanding is that we will unify not only exporting to zip for both Docker and Cloud deployment but also the `VespaDocker.export_application_files` method, otherwise we will have code duplication that will be hard to support going forward.

vespa/deployment.py
```diff
@@ -110,30 +115,49 @@ def _run_vespa_engine_container(
...
-                    self.container_image,
-                    detach=True,
-                    mem_limit=container_memory,
-                    name=application_name,
-                    hostname=application_name,
-                    privileged=True,
-                    volumes={disk_folder: {"bind": "/app", "mode": "rw"}},
-                    ports={8080: self.local_port},
-                )
+                if disk_folder:
```
thigm85, line 118: My understanding is that this option will be removed when deploying through zip is validated.

vespa/deployment.py
```diff
@@ -307,10 +357,9 @@ def deploy(
         :return: a Vespa connection instance.
         """
         if not self.disk_folder:
-            self.disk_folder = os.path.join(os.getcwd(), application_package.name)
+            return self._execute_deployment_zip(app_package=application_package)
```
thigm85, line 360: It would be nice to add an integration test in this PR that uses the functionality implemented here, namely `_execute_deployment_zip` which is triggered when not specifying `disk_folder`.

thigm85 (changes requested):

vespa/package.py
kkraune: Yes I am working on the export, too. In the transition it will be duplication in order to not break too much / there is so much code to change. When done and no tests fail, we can look at notebooks and change these - then do final cleanup

vespa/deployment.py
kkraune: correct

vespa/deployment.py
kkraune: agreed, will do

kkraune:
PTAL at your convenience

thigm85 (approved):

thigm85:
I approved the changes but there is one notebook failure due to a lack of the `disk_folder` variable.

kkraune:
I see the error now in https://cd.screwdriver.cd/pipelines/7055/builds/785923/steps/run-notebooks-except-cloud-related, I'll fix it in next PR

PR #320, opened 2022-04-22
Kkraune/misc

kkraune:
- no functional changes, just renames / cleanup for less warnings


kkraune:
@freva please review - should be no functional changes except for the contant -> constant typo, the correction is correct per @thigm85 (who is on PTO these days). This is just cleaning up warnings in the IDE

freva (commented):
Some methods are changed to static, but their are still invoked via `self`...

thigm85:
@kkraune I will be able to review this today as well.

thigm85 (approved):

freva (commented):
I think you also need to update `return model.parse_vespa_prediction(` in `vespa/application.py` to `TextTask.parse_vespa_prediction(` (Ref static method change in `vespa/ml.py`)

kkraune:
> model.parse_vespa_prediction

you are right! thanks

freva (approved):

PR #324, opened 2022-04-27
Use a common requirements section

kkraune:
- simplify the notebooks by referring to common install steps in the docs


kkraune:
linkcheck fails on a link that is also fixed in this PR, but not live yet, so OK

thigm85 (approved):

PR #326, opened 2022-05-01
Print error message at deploy error

kkraune:
A deployment error will now look like:

````
RuntimeError: Deployment failed, code: 400, message: {'error-code': 'BAD_REQUEST', 'message': 'Error preprocessing application package for default.default, session 9: Error preprocessing /opt/vespa/var/db/vespa/config_server/serverdb/tenants/default/sessions/9/services.xml: The element type "ccontainer" must be terminated by the matching end-tag "</ccontainer>".: The element type "ccontainer" must be terminated by the matching end-tag "</ccontainer>".'}
````


thigm85 (approved):
Much better, thanks.

PR #361, opened 2022-07-06
Bug fix for feed_batch

msminhas93:
Fixed bug in feed_batch by replacing 0 return with the expected feed_results variable

### DIFF
--- a/vespa/application.py
+++ b/vespa/application.py
@@ -521,6 +521,7 @@ class Vespa(object):
         mini_batches = [
             batch[i : i + batch_size] for i in range(0, len(batch), batch_size)
         ]
+        batch_http_responses = []
         for idx, mini_batch in enumerate(mini_batches):
             feed_results = self._feed_batch(
                 batch=mini_batch,
@@ -530,6 +531,7 @@ class Vespa(object):
                 total_timeout=total_timeout,
                 namespace=namespace,
             )
+            batch_http_responses.extend(feed_results)
             status_code_summary = Counter([x.status_code for x in feed_results])
             print(
                 "Successful documents fed: {}/{}.\nBatch progress: {}/{}.".format(
@@ -539,7 +541,7 @@ class Vespa(object):
                     len(mini_batches),
                 )
             )
-        return feed_results
+        return batch_http_responses
 
     def feed_df(self, df: DataFrame, include_id: bool = True, **kwargs):
         """
--- a/vespa/application.py
+++ b/vespa/application.py
@@ -539,7 +539,7 @@ class Vespa(object):
                     len(mini_batches),
                 )
             )
-        return 0
+        return feed_results
 
     def feed_df(self, df: DataFrame, include_id: bool = True, **kwargs):
         """


thigm85:
Hi @msminhas93, thanks for the PR. 

The problem with this fix is that starting on version [0.20.0](https://github.com/vespa-engine/pyvespa/releases/tag/v0.20.0) feed_batch works by sending mini-batches to the Vespa application. So, returning `feed_results` will only return the results of the last mini-batch. The reason to send mini-batches is to make the method more reliable when feeding a large number of documents.

A possible solution would be to allow the parameter `batch_size` to be `None` in which case the method sends all the documents at once and return `feed_results`. What do you think?

thigm85 (changes requested):
Hi @msminhas93, thanks for the PR.

The problem with this fix is that starting on version [0.20.0](https://github.com/vespa-engine/pyvespa/releases/tag/v0.20.0) feed_batch works by sending mini-batches to the Vespa application. So, returning feed_results will only return the results of the last mini-batch. The reason to send mini-batches is to make the method more reliable when feeding a large number of documents.

A possible solution would be to allow the parameter batch_size to be None in which case the method sends all the documents at once and return feed_results. What do you think?

msminhas93:
Hey @thigm85, thank you for the prompt response!

Sorry I missed looking at the batching step since for my test case the input batch was of size 10 and I was getting the same number of responses in the feed_results (because batch_size>input_batch_size).

In that case, do you think just having another list to keep track of the HTTP responses and returning that would make sense? That would allow it to work with batching not equal to None too.

```
    def feed_batch(
        self,
        batch: List[Dict],
        schema: Optional[str] = None,
        asynchronous=True,
        connections: Optional[int] = 100,
        total_timeout: int = 100,
        namespace: Optional[str] = None,
        batch_size=1000,
    ):
        """
        Feed a batch of data to a Vespa app.

        :param batch: A list of dict containing the keys 'id' and 'fields' to be used in the :func:`feed_data_point`.
        :param schema: The schema that we are sending data to. The schema is optional in case it is possible to infer
            the schema from the application package.
        :param asynchronous: Set True to send data in async mode. Default to True.
        :param connections: Number of allowed concurrent connections, valid only if `asynchronous=True`.
        :param total_timeout: Total timeout in secs for each of the concurrent requests when using `asynchronous=True`.
        :param namespace: The namespace that we are sending data to. If no namespace is provided the schema is used.
        :param batch_size: The number of documents to feed per batch.
        :return: List of HTTP POST responses
        """
        mini_batches = [
            batch[i : i + batch_size] for i in range(0, len(batch), batch_size)
        ]
        batch_post_responses = []
        for idx, mini_batch in enumerate(mini_batches):
            feed_results = self._feed_batch(
                batch=mini_batch,
                schema=schema,
                asynchronous=asynchronous,
                connections=connections,
                total_timeout=total_timeout,
                namespace=namespace,
            )
            batch_post_responses.extend(feed_results)
            status_code_summary = Counter([x.status_code for x in feed_results])
            print(
                "Successful documents fed: {}/{}.\nBatch progress: {}/{}.".format(
                    status_code_summary[200],
                    len(mini_batch),
                    idx + 1,
                    len(mini_batches),
                )
            )
        return batch_post_responses
```

thigm85:
The disadvantage of that approach is that it will increase memory usage in case you want to send a lot of data. But I think it makes sense. If the amount of data is too large the user can make multiple feed_batch calls.

msminhas93:
> The disadvantage of that approach is that it will increase memory usage in case you want to send a lot of data. But I think it makes sense. If the amount of data is too large the user can make multiple feed_batch calls.

Well then instead of returning the entire HTTP object, we can also choose to just return the status code if we want and then modify the doc to reflect that. Since the status code should be enough to verify the success/failure. Do you have any preference? I can then update the PR accordingly.

thigm85:
Let's keep the entire HTTP object to be consistent with the other document operations. I can review the PR with the modification.

thigm85 (approved):
LGTM

PR #379, opened 2022-09-26
Make it easier to use existing tenant/app/instance

kkraune:


thigm85 (approved):
LGTM

thigm85:
@kkraune The link checker is running the notebook because none of the cells have been run by you, and the notebook is not executable without user intervention. So I suggest you run at least some of the cells and save the notebook. The link checker assumes the notebook was already run and skips this step.

thigm85 (changes requested):
Run at least one notebook cell to avoid it being run by the link-checker.

kkraune:
Thanks! I did not know about the link checker / run cells. will this do?

thigm85 (approved):
Thanks

PR #382, opened 2022-10-08
Move query module to learntorank

thigm85:
* Moved query module (QueryModel concept) from pyvespa to learntorank
* app.query and app.query_batch will now only accept YQL via the body parameter (lots of tests to rewrite and move to learntorank)
* Moved app.collect_vespa_features and app.store_vespa features to learntorank
* Removed gallery module from pyvespa


kkraune (approved):
Great!

PR #398, opened 2022-11-11
use input name input

kkraune:
- preliminary fix: it seems like it expects "input" - persisting this in a PR until I understand more

FYI @lesters

### DIFF
--- a/docs/sphinx/source/use_cases/tensorflow-via-onnx.ipynb
+++ b/docs/sphinx/source/use_cases/tensorflow-via-onnx.ipynb
@@ -858,7 +858,7 @@
     "    OnnxModel(\n",
     "        model_name=\"ltr_tensorflow\",\n",
     "        model_file_path=\"simpler_keras_model.onnx\",\n",
-    "        inputs={\"input:0\": \"vespa_input\"},\n",
+    "        inputs={\"input\": \"vespa_input\"},\n",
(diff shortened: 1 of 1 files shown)


thigm85:
@kkraune Is this related to the breaking notebook in the pipeline? I talked with Jon, and it seems there were Vespa changes related to moving ONNX models back into the config model now that TensorFlow support is gone.

bratseth:
After actually looking at it I guess it's not caused by that, but I also don't see what it could be related to.
Looks like there is no way in screwdriver to see older failures ... does any of you know when this started to fail?

thigm85:
@bratseth it started to fail with this commit https://github.com/vespa-engine/pyvespa/commit/6e80f1ab9ab741c3640c037f128c8e501767f8c4 (which make unrelated changes AFAIK). This is the [offending SD build details](https://cd.screwdriver.cd/pipelines/7055/builds/830802/steps/run-notebooks-except-cloud-related). You can see this was on 08/11/2022.

thigm85:
But it seems that the changes made by @kkraune in this PR worked, so it seems the problem originated from changes in the ONNX export of the model into .onnx file. So, not related to Vespa. Case closed?

thigm85 (approved):
LGTM

bratseth:
Hm, yes.

PR #399, opened 2022-12-07
Adding the possibility to debug the docker image

tkaessmann:
Hey Vespa Team,
What do you think about the idea to add the possiblity to debug the local docker image?
In our case this would be very helpful, to easily spin up a local vespa and fill it with some dummy data to test a little bit with real data.


kkraune (approved):
Thanks!

kkraune:
(I'll add a note in the docs somewhere)

keep the PRs coming :-)

PR #401, opened 2022-12-12
add py.typed to the package

maxice8:
This is required so mypy can import this package to perform type-checking, I see some functions are missing Return annotations, if required I can update this PR so it does that too.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -37,6 +37,7 @@ setuptools.setup(
     },
     python_requires=">=3.6",
     zip_safe=False,
+    package_data={"vespa": ["py.typed"]},
     data_files=[
         (
             "templates",
--- a/vespa/application.py
+++ b/vespa/application.py
@@ -167,6 +167,7 @@ class Vespa(object):
         Wait for application ready.
 
         :param max_wait: Seconds to wait for the application endpoint
+        :raises RuntimeError: If the application did not start in the time given by the :max_wait: param.
         :return:
         """
         try_interval = 5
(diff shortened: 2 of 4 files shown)


bratseth (approved):
That would be great - thanks.

maxice8:
I have added a "few" (7) commits that should help annotation and minor things I found, I think the commit titles should be self-explanatory enough but for the sake of completeness:

- Added `-> None` as the return type for functions that return nothing
- Narrowed `disk_folder` in `_to_application_zip` as its usage with a `"".format()` makes it so only str (and LiteralString) can be used
- Made `max_wait` an int, this might be breaking if someone uses `wait_for_config_server_start(4.5)` or another float value, while using a float value just means you're using an int rounded up (4.5 will wait for 5 seconds) it might cause annoyance to already existing users.
- Replaced usages of bare `(tuple_elem_1, tuple_elem_2, ...)` with the recommended `Tuple[tuple_elem_1, tuple_elem_2, ...]`
- Annotated the `__enter__` dunder to return its object ("VespaCloud" in this case).
- Annotated `__repr__` to return a str
- Made `__eq__` return `NotImplemented` (see: https://docs.python.org/3/library/constants.html#NotImplemented for behaviour) when `other` is not of the same object as self, also annotated `other` as an `object`, per guidelines.

maxice8:
I have gone over the other python files (that are not tests) and fixed the most obvious/simplest errors.

bratseth:
Great!

PR #402, opened 2022-12-12
More thoroughly document  {start,stop,restart}_services() methods

maxice8:
This documents the RuntimeError exceptions that are raised in each method, and it also shortly describes how each of them operates.


kkraune (approved):
Thanks!

PR #403, opened 2022-12-12
add support for match in Field

maxice8:
This adds supports for match blocks as described here: https://docs.vespa.ai/en/reference/schema-reference.html#match.

The new `match` optional attribute can use a simple `str` for any property that does not require a value (like `exact` and `gram`), and uses a `Tuple[str, str]` for properties that require a value (like `exact-terminator` and `gram-size`).=

Example Field:

```python3
Field(
	name="abstract",
	type="string",
	match=["exact", ("exact-terminator", '"@%"'), "gram", ("gram-size", 3)],
),
```

leads to this generated schema:

```
field abstract type string {
    match {
        exact
        exact-terminator: "@%"
        gram
        gram-size: 3
    }
}
```

### DIFF
--- a/vespa/package.py
+++ b/vespa/package.py
@@ -3,7 +3,7 @@ import zipfile
 
 from pathlib import Path
 from shutil import copyfile
-from typing import List, Optional, Union, Dict
+from typing import List, Optional, Tuple, Union, Dict
 from collections import OrderedDict
 from jinja2 import Environment, PackageLoader, select_autoescape
 from io import BytesIO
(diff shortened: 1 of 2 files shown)


maxice8:
I am investigating these errors :sweat_smile: unfortunately I could not get Screwdriver to work locally to run the tests.

maxice8:
Should be good to review, issues were Doctests in other parts of the code and a typo in one doctest.

lesters (approved):
LGTM!

PR #405, opened 2022-12-20
Add Summary from the Schema.

maxice8:
Created using the [Schema reference](https://docs.vespa.ai/en/reference/schema-reference.html#summary).

I am not fully happy with this, specially that it cannot output simple `summary: [property]`, instead doing `summary {\n	[property]\n}`, but otherwise it should be able to output anything you throw at it.

The expectation is being able to add `document-summary` at a later date, which this is a requirement.

### DIFF
--- a/vespa/package.py
+++ b/vespa/package.py
@@ -129,6 +129,7 @@ class Field(object):
         match: Optional[List[Union[str, Tuple[str, str]]]] = None,
         weight: Optional[int] = None,
         bolding: Optional[Literal[True]] = None,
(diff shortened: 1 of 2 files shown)


maxice8:
This has been rebased ontop of the recent additions of `weight` and `bolding` from #406

maxice8:
Should be ready for review, please let me know if anything needs changed to fit better with Vespa/be more intuitive for the user.

lesters (approved):
Great, LGTM! 👍

PR #406, opened 2022-12-21
add weight and bolding support for Field

maxice8:
This adds supports for the weight and bolding fields, as described [here](https://docs.vespa.ai/en/reference/schema-reference.html#bolding) and [here](https://docs.vespa.ai/en/reference/schema-reference.html#bolding).


bratseth (approved):
LGTM!

PR #408, opened 2023-01-06
Implement DocumentSummary

maxice8:
This implements DocummentSummary, as described in
https://docs.vespa.ai/en/reference/schema-reference.html#document-summary,
making use of the previous Summary work from #405

It also changes how `Summary` is generated in the final schema by adding a new property `as_lines`, which represents each line of summary class as it would be written out in the final file on disk.

This has the advantage of moving all the logic to python which is much easier to handle than Jinja, this has the effect of allowing us to do single-attribute summaries like `summary: dynamic`, and also we can handle Summaries with no body (like `summary simple type string {}`).

The one drawback I see from this `as_lines` approach is that the indent is tied to the python code, in Jinja we could write it out the fields and the structure would lead to the indentation, but with `as_lines` we add 4 whitespaces to lines that are in the body of the summary.

### DIFF
--- a/vespa/package.py
+++ b/vespa/package.py
@@ -387,6 +387,77 @@ class Struct(object):
         )
 
 
+class DocumentSummary(object):
+    def __init__(
+        self,
+        name: str,
+        inherits: Optional[str] = None,
+        summary_fields: Optional[List[Summary]] = None,
+        from_disk: Optional[Literal[True]] = None,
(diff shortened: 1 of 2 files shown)


lesters (approved):
LGTM!

maxice8:
Rebased on top of latest changes to master

PR #410, opened 2023-01-09
Add Stemming support for fields

maxice8:
This allows fields to define a Stemming as any string


lesters (approved):
👍

maxice8:
Considering https://github.com/vespa-engine/pyvespa/pull/412#discussion_r1066205019 I think we should hold on from adding anything to the Fields constructor until we sort out how to prevent it from greatly increasing in quantity

PR #411, opened 2023-01-10
rank in field

maxice8:
- Add rank= in field
- Add type-hints to dunder of :class:`Field`
- update doctests for :class:`Document`


lesters (approved):
LGTM, but again, constructor should be addressed.

PR #412, opened 2023-01-10
Implement query-command for field

maxice8:
This implements query-command inside a :class:`Field`

### DIFF
--- a/vespa/package.py
+++ b/vespa/package.py
@@ -130,6 +130,7 @@ class Field(object):
         weight: Optional[int] = None,
         bolding: Optional[Literal[True]] = None,
         summary: Optional[Summary] = None,
+        query_command: Optional[List[str]] = None,
     ) -> None:
         """
         Create a Vespa field.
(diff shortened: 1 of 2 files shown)


vespa/package.py
```diff
@@ -130,6 +130,7 @@ def __init__(
         weight: Optional[int] = None,
         bolding: Optional[Literal[True]] = None,
         summary: Optional[Summary] = None,
+        query_command: Optional[List[str]] = None,
     ) -> None:
```
lesters, line 134: This constructor is getting out of hand. I wonder if this should be changed to a `kwargs` or a simple dict?

lesters (approved):
See comment about constructor. Other than that, LGTM.

vespa/package.py
maxice8: Yeah, it is filling up fast, I think we can have it accept a `TypedDict` that has `NotRequired[Type]` in all fields so you can pass a dictionary of what you want to configure.

Something like (writing off the top of my head):

```python3
class FieldConfigurationDict(TypedDict):
	weight: NotRequired[int]
	bolding: NotRequired[Literal[True]]
	query_command: NotRequired[List[str]]

class Field:
	def __init__(self, all, previous, constructors, to keep, backwards, compat, **kwargs: FieldConfigurationDict):
		"""Code that runs over every value and assigns it correctly"""
```

PR #413, opened 2023-01-10
Implement Struct Field, and allow Field to use it

maxice8:
- Implement StructField
- Add support for StructField inside Field

### DIFF
--- a/vespa/package.py
+++ b/vespa/package.py
@@ -217,6 +217,7 @@ class FieldConfiguration(TypedDict, total=False):
     stemming: str
     rank: str
     query_command: str
+    struct_fields: List[StructField]
 
 
 class Field(object):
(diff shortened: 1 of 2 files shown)


vespa/package.py
```diff
@@ -247,7 +338,8 @@ def __repr__(self):
             repr(self.match),
             repr(self.weight),
             repr(self.bolding),
-            repr(self.summary)
+            repr(self.summary),
+            repr(self.struct_fields),
```
lesters, line 342: I guess this PR conflicts with https://github.com/vespa-engine/pyvespa/pull/411 and https://github.com/vespa-engine/pyvespa/pull/412 and https://github.com/vespa-engine/pyvespa/pull/410 here?

lesters (approved):
LGTM, but I think some work needs to be done before merging due to the other PRs wirking on fields.

vespa/package.py
maxice8: Yes, each of my PRs (that the Field class will conflict upon merging either of them), in any case I think we should deal with the [amount of constructors issue](https://github.com/vespa-engine/pyvespa/pull/412#discussion_r1066205019) before merging any of them.

maxice8:
Rebased on top of master

PR #414, opened 2023-01-11
Use **kwargs instead of defining every possible field for Field's __init__

maxice8:
This replaces the definition of every single parameter into a separate TypedDict, based on this [comment](https://github.com/vespa-engine/pyvespa/pull/412#discussion_r1066210333).

```python3
class FieldConfiguration(TypedDict, total=False):
    indexing: List[str]
    attribute: List[str]
    ann: HNSW
    match: List[Union[str, Tuple[str, str]]]
    weight: int
    bolding: Literal[True]
    summary: Summary
```

All current accepted fields are there, and any new fields like the ones in #410 and #411 can only be added there. They would then be documented as `:key foo:` instead of `:param foo:`.

The `__init__` would then accept `**kwargs: Unpack[FieldConfiguration]` like so:

```python3
def __init__(
        self,
        name: str,
        type: str,
        indexing: Optional[List[str]] = None,
        index: Optional[str] = None,
        attribute: Optional[List[str]] = None,
        ann: Optional[HNSW] = None,
        match: Optional[List[Union[str, Tuple[str, str]]]] = None,
        weight: Optional[int] = None,
        bolding: Optional[Literal[True]] = None,
        summary: Optional[Summary] = None,
        **kwargs: Unpack[FieldConfiguration],
    ) -> None:
```

The current `__init__` retains the current parameters for backwards compatibility, it is reasonable that someone might have done `Field("name", "type", ["attribute"])` and replacing it with only `**kwargs` would break it.

The assignment is now done via `kwargs.get("attr", DefaultValue)`, with `DefaultValue` falling back on the parameter if the parameter already existed at the time of introduction, or the value itself if not (See below for example).

```python3
self.name = name
self.type = type
self.indexing = kwargs.get("indexing", indexing)
self.attribute = kwargs.get("attribute", attribute)
self.index = kwargs.get("index", index)
self.ann = kwargs.get("ann", ann)
self.match = kwargs.get("match", match)
self.weight = kwargs.get("weight", weight)
self.bolding = kwargs.get("bolding", bolding)
self.summary = kwargs.get("summary", summary)
# Hypothethical new value
# Note the bare type instead of falling back on the parameter
self.stemming = kwargs.get("stemming", None)
```

cc: @lesters


maxice8:
Old python doesn't have Unpack in `typing`, only `typing_extensions` so I have added a workaround, I don't know if Vespa's minimum required version has it.

maxice8:
Another option it to make use of https://docs.python.org/3/library/dataclasses.html.

Here is the `Field` class using `@dataclass`

```python3
from dataclasses import dataclasses

@dataclass(init=True, eq=True)
class Field(object):

    """
    Create a Vespa field.

    Check the `Vespa documentation <https://docs.vespa.ai/en/reference/schema-reference.html#field>`__
    for more detailed information about fields.

    Once we have an :class:`ApplicationPackage` instance containing a :class:`Schema` and a :class:`Document`,
    we usually want to add fields so that we can store our data in a structured manner.
    We can accomplish that by creating :class:`Field` instances
    and adding those to the :class:`ApplicationPackage` instance via :class:`Schema` and :class:`Document` methods.

    :param name: Field name.
    :param type: Field data type.
    :param indexing: Configures how to process data of a field during indexing.
    :param index: Sets index parameters. Content in fields with index are normalized and tokenized by default.
    :param attribute:  Specifies a property of an index structure attribute.
    :param ann: Add configuration for approximate nearest neighbor.
    :param match: Set properties that decide how the matching method for this field operate.
    :param weight: Sets the weight of the field, using when calculating Rank Scores.
    :param bolding: Whether to highlight matching query terms in the summary.
    :param summary: Add configuration for summary of the field.

    >>> Field(name = "title", type = "string", indexing = ["index", "summary"], index = "enable-bm25")
    Field('title', 'string', ['index', 'summary'], 'enable-bm25', None, None, None, None, None, None)

    >>> Field(
    ...     name = "abstract",
    ...     type = "string",
    ...     indexing = ["attribute"],
    ...     attribute=["fast-search", "fast-access"]
    ... )
    Field('abstract', 'string', ['attribute'], None, ['fast-search', 'fast-access'], None, None, None, None, None)

    >>> Field(name="tensor_field",
    ...     type="tensor<float>(x[128])",
    ...     indexing=["attribute"],
    ...     ann=HNSW(
    ...         distance_metric="euclidean",
    ...         max_links_per_node=16,
    ...         neighbors_to_explore_at_insert=200,
    ...     ),
    ... )
    Field('tensor_field', 'tensor<float>(x[128])', ['attribute'], None, None, HNSW('euclidean', 16, 200), None, None, None, None)

    >>> Field(
    ...     name = "abstract",
    ...     type = "string",
    ...     match = ["exact", ("exact-terminator", '"@%"',)],
    ... )
    Field('abstract', 'string', None, None, None, None, ['exact', ('exact-terminator', '"@%"')], None, None, None)

    >>> Field(
    ...     name = "abstract",
    ...     type = "string",
    ...     weight = 200,
    ... )
    Field('abstract', 'string', None, None, None, None, None, 200, None, None)

    >>> Field(
    ...     name = "abstract",
    ...     type = "string",
    ...     bolding = True,
    ... )
    Field('abstract', 'string', None, None, None, None, None, None, True, None)

    >>> Field(
    ...     name = "abstract",
    ...     type = "string",
    ...     summary = Summary(None, None, ["dynamic", ["bolding", "on"]]),
    ... )
    Field('abstract', 'string', None, None, None, None, None, None, None, Summary(None, None, ['dynamic', ['bolding', 'on']]))
    """
    name: str
    type: str
    indexing: Optional[List[str]] = None
    index: Optional[str] = None
    attribute: Optional[List[str]] = None
    ann: Optional[HNSW] = None
    match: Optional[List[Union[str, Tuple[str, str]]]] = None
    weight: Optional[int] = None
    bolding: Optional[Literal[True]] = None
    summary: Optional[Summary] = None

    @property
    def indexing_to_text(self) -> Optional[str]:
        if self.indexing is not None:
            return " | ".join(self.indexing)

    def __repr__(self):
        return "{0}({1}, {2}, {3}, {4}, {5}, {6}, {7}, {8}, {9}, {10})".format(
            self.__class__.__name__,
            repr(self.name),
            repr(self.type),
            repr(self.indexing),
            repr(self.index),
            repr(self.attribute),
            repr(self.ann),
            repr(self.match),
            repr(self.weight),
            repr(self.bolding),
            repr(self.summary)
        )
```

Note that you can also replace the `__repr__` if you use `repr=True` in the decorator call, **BUT** it will make the repr use the parameter name like this:

```python3
Field("abstract", "string")
Field(name='abstract', type='string', indexing=None)
```

maxice8:
Ping, this needs to be solved before any new features can be added to :class:Field

lesters:
@maxice8 I think the kwargs method is a reasonable way of doing this. Let's go forward with this.

lesters (approved):
LGTM

maxice8:
Sounds good, enabled review so it can be merged.

PR #421, opened 2023-01-25
Added ApplicationConfiguration to set generic configurations for application

JeanExtreme002:

### DIFF
--- a/vespa/package.py
+++ b/vespa/package.py
@@ -1523,7 +1523,7 @@ class ApplicationPackage(object):
         return schema_template.render(
             application_name=self.name,
             schemas=self.schemas,
-            user_version = self.user_version,
+            user_version=self.user_version,
             stateless_model_evaluation=self.stateless_model_evaluation,
         )
 
--- a/vespa/package.py
+++ b/vespa/package.py
@@ -1523,7 +1523,7 @@ class ApplicationPackage(object):
         return schema_template.render(
             application_name=self.name,
             schemas=self.schemas,
-            user_version=self.user_version,
+            user_version = self.user_version,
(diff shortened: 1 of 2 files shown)


lesters (changes requested):
Thanks for this! I feel however that this parameter should not be a top-level argument to the constructor, as this sets a certain configuration setting. There could be many such. What do you think about making this a bit more general, that can support any config setting?

JeanExtreme002:
@lesters Done! The only thing I'll do now is fixing the `services_to_text` text fomatting, because for some reason `</config>` is going out from indentation. But everything is already working! :smile:

lesters (commented):
Looks good, but there are test errors?

JeanLoui-Escavador:
> Looks good, but there are test errors?

Other pull requests had that same problem on Screwdriver/7055/PR:unit-tests

vespa/package.py
```diff
@@ -1370,6 +1370,48 @@ def __repr__(self):
...
+
+        :param name: Configuration name.
+        :param value: Either a string or a Dict (it may be a nested dict) of values.
+
+        Example:
+        >>> ApplicationConfiguration(
+        >>>     name="container.handler.observability.application-userdata",
+        >>>     value={"version": "my-version"}
+        >>> )
+        """
```
lesters, line 1389: Unit test fail is related to this. The doctesting fails here. Should be "..." on these lines?

lesters (changes requested):
Looks ok, but please fix unit test failures.

JeanExtreme002:
Sorry for my mistake. It's done! @lesters

lesters (approved):
LGTM!

PR #430, opened 2023-02-06
Kkraune/refactor tests

kkraune:
- FYI @jobergum 
- Less deps / work around problem for now with failing Docker stop


kkraune:
CC @lesters

PR #432, opened 2023-02-07
Test only msmarco in cloud

kkraune:
- the _functionality_ or the remove apps here are tested in the non-cloud test suite
- no need to run the same tests over again using Vespa Cloud - this will cut test time

FYI @jobergum


lesters (approved):
👍

PR #447, opened 2023-02-10
Revert back - link check fail

kkraune:
- revert - sphinx error, don't understand why - do back to a working version


kkraune:
removed the required cells and unit tests run, so good now :-)

jobergum (approved):

PR #491, opened 2023-03-22
Remove extras_require

kkraune:
- `python3 -m pip install -e .[full]` pulls in this, and it is not needed now, as code is moved to learntorank


kkraune:
review only, lets see the test runs first

lesters (approved):

PR #466, opened 2023-02-16
Kkraune/exceptions

kkraune:
- It is a problem that errors are hidden, making it hard to understand why examples/notebooks don't work
- I think it is better to throw an exception on 4xx/5xx - this is a suggestion, if we decide to do it this way, must do in more places - for discussion

Example:
````
app.feed_data_point(schema="sentence", data_id=idx, fields=sentence)
````
will now fail on disk full:
````
---------------------------------------------------------------------------
HTTPError                                 Traceback (most recent call last)
Cell In [18], line 2
      1 for context in context_data:
----> 2     app.feed_data_point(schema="context", data_id=context["context_id"], fields=context)

File ~/github/vespa-engine/pyvespa/vespa/application.py:298, in Vespa.feed_data_point(self, schema, data_id, fields, namespace)
    295     namespace = schema
    297 with VespaSync(app=self) as sync_app:
--> 298     return sync_app.feed_data_point(
    299         schema=schema, data_id=data_id, fields=fields, namespace=namespace
    300     )

File ~/github/vespa-engine/pyvespa/vespa/application.py:834, in VespaSync.feed_data_point(self, schema, data_id, fields, namespace)
    832 vespa_format = {"fields": fields}
    833 response = self.http_session.post(end_point, json=vespa_format, cert=self.cert)
--> 834 response.raise_for_status()
    835 return VespaResponse(
    836     json=response.json(),
    837     status_code=response.status_code,
    838     url=str(response.url),
    839     operation_type="feed",
    840 )

File /usr/local/Cellar/jupyterlab/3.4.8/libexec/lib/python3.10/site-packages/requests/models.py:1021, in Response.raise_for_status(self)
   1016     http_error_msg = (
   1017         f"{self.status_code} Server Error: {reason} for url: {self.url}"
   1018     )
   1020 if http_error_msg:
-> 1021     raise HTTPError(http_error_msg, response=self)

HTTPError: 507 Server Error: Insufficient Storage for url: http://localhost:8080/document/v1/context/context/docid/0
````

PR #494, opened 2023-04-09
Prevent errors from being raised when calling APIs

rejasupotaro:
## Summary

This PR prevents errors from being raised when calling Vespa APIs.

It is just one of the possible options for the problems I have encountered.

## Context

I'm new to Vespa. I attempted to call the search API, but the following code resulted in the error below.

```python
body = {
    "yql": "select * from sources typo where userQuery()",
    "query": "hello",
    "type": "any",
    "ranking": "random",
    "hits": 10,
}
vespa_app.query(body)
```

```python
requests.exceptions.HTTPError: 400 Client Error: Bad Request for url: http://localhost:8080/search/
```

I was struggling with debugging because the error message didn't provide enough details, but I managed to find out that there was a typo in `sources typo`.

## Issue

Reading the code, `raise_for_status` is called to raise an error when the HTTP status code is not successful.

```python
# vespa/application.py
    def query(
        self,
        body: Optional[Dict] = None,
    ) -> VespaQueryResponse:
        ...
        response = self.http_session.post(self.app.search_end_point, json=body, cert=self.cert)
        response.raise_for_status()
        return VespaQueryResponse(
            json=response.json(), status_code=response.status_code, url=str(response.url)
        )
```

It seems that this was added in https://github.com/vespa-engine/pyvespa/pull/466 This PR enabled us to receive a forced notification when an error occurred. 

<details>

```python
# requests/models.py
    def raise_for_status(self):
        """Raises :class:`HTTPError`, if one occurred."""

        http_error_msg = ""
        if isinstance(self.reason, bytes):
            # We attempt to decode utf-8 first because some servers
            # choose to localize their reason strings. If the string
            # isn't utf-8, we fall back to iso-8859-1 for all other
            # encodings. (See PR #3538)
            try:
                reason = self.reason.decode("utf-8")
            except UnicodeDecodeError:
                reason = self.reason.decode("iso-8859-1")
        else:
            reason = self.reason

        if 400 <= self.status_code < 500:
            http_error_msg = (
                f"{self.status_code} Client Error: {reason} for url: {self.url}"
            )

        elif 500 <= self.status_code < 600:
            http_error_msg = (
                f"{self.status_code} Server Error: {reason} for url: {self.url}"
            )

        if http_error_msg:
            raise HTTPError(http_error_msg, response=self)
```

</details>

However, the error doesn't include a message from Vespa. The message that should have been returned was as follows:

```python
{
  'root': {
    'id': 'toplevel',
    'relevance': 1.0,
    'fields': {
      'totalCount': 0
    },
    'errors': [
      {
        'code': 4,
        'summary': 'Invalid query parameter',
        'message': "Could not resolve source ref 'typo'. Valid source refs are amazon_content."
      }
    ]
  }
}
```

I would be able to identify the cause instantly if I could see the message.

## Options

There are several approaches that I think we can resolve this issue.

### 1. Stop calling `raise_for_status`

This is the approach I took in this PR. Errors are no longer raised, and it is now the responsibility of the caller to decide how to handle errors by checking the status code.

### 2. Raise a custom error

If we want to raise an error and include a message from Vespa, it would be better to create a custom error rather than raising `HTTPError`.

## Example of [elasticsearch-py](https://github.com/elastic/elasticsearch-py)

The Bulk API has the [`raise_on_error`](https://github.com/elastic/elasticsearch-py/blob/main/elasticsearch/helpers/actions.py#L363) param. If True, [an error is raised](https://github.com/elastic/elasticsearch-py/blob/f9df3af3db99e6ac4db5cfed22f819469c4b41fb/elasticsearch/helpers/actions.py#L307-L310).

### DIFF
--- a/vespa/application.py
+++ b/vespa/application.py
@@ -825,7 +825,6 @@ class VespaSync(object):
         :param fields: Dict containing all the fields required by the `schema`.
         :param namespace: The namespace that we are sending data to. If no namespace is provided the schema is used.
         :return: Response of the HTTP POST request.
-        :raises HTTPError: if one occurred
         """
 
         if not namespace:
@@ -836,7 +835,6 @@ class VespaSync(object):
         )
         vespa_format = {"fields": fields}
         response = self.http_session.post(end_point, json=vespa_format, cert=self.cert)
-        response.raise_for_status()
         return VespaResponse(
             json=response.json(),
             status_code=response.status_code,
@@ -855,10 +853,8 @@ class VespaSync(object):
 
         :param body: Dict containing all the request parameters.
         :return: Either the request body if debug_request is True or the result from the Vespa application
-        :raises HTTPError: if one occurred
         """
         response = self.http_session.post(self.app.search_end_point, json=body, cert=self.cert)
-        response.raise_for_status()
         return VespaQueryResponse(
             json=response.json(), status_code=response.status_code, url=str(response.url)
         )
@@ -873,7 +869,6 @@ class VespaSync(object):
         :param data_id: Unique id associated with this data point.
         :param namespace: The namespace that we are deleting data from.
         :return: Response of the HTTP DELETE request.
-        :raises HTTPError: if one occurred
         """
         if not namespace:
             namespace = schema
@@ -882,7 +877,6 @@ class VespaSync(object):
             self.app.end_point, namespace, schema, str(data_id)
         )
         response = self.http_session.delete(end_point, cert=self.cert)
-        response.raise_for_status()
         return VespaResponse(
             json=response.json(),
             status_code=response.status_code,
@@ -900,7 +894,6 @@ class VespaSync(object):
         :param schema: The schema that we are deleting data from.
         :param namespace: The namespace that we are deleting data from.
         :return: Response of the HTTP DELETE request.
-        :raises HTTPError: if one occurred
         """
         if not namespace:
             namespace = schema
@@ -909,7 +902,6 @@ class VespaSync(object):
             self.app.end_point, namespace, schema, content_cluster_name
         )
         response = self.http_session.delete(end_point, cert=self.cert)
-        response.raise_for_status()
         return response
 
     def get_data(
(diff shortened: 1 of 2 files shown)


kkraune:
This is great feedback! Yes, we added the exception to avoid silent failures (a lot of the examples have little error checking for brevity), but I agree this is quite coarse - I suggest we improve it instead :-) It is much better if the Vespa error message is printed 

https://docs.vespa.ai/en/reference/query-api-reference.html#http-status-codes

I suggest we create a little utility function to check for errors, emit a more meaningful error message - and raise a proper error (HTTP or Vespa error)

rejasupotaro:
@kkraune Thank you for your feedback! In fact, I would prefer to raise an error with an appropriate message.

I think we need to consider how to merge the information from different sources:

- `requests.HTTPError`, which contains `status_code`, `reason`, and `url`
- `vespa.exceptions.VespaError` (tentative), which can have multiple errors and each error has its own `code` (!= status code), `summary`, and `message`.

I will create another PR to keep things organized.

rejasupotaro:
Here is the PR: https://github.com/vespa-engine/pyvespa/pull/498 I think they each have their own advantages and disadvantages. In particular, there are differences in error handling as shown below.

### Prevent errors from being raised when calling APIs

It's now up to the caller to decide how to handle it. They can either choose to ignore the errors or handle them appropriately.

```python
# In my app
response = vespa_app.query(body)
match response.status_code:
    case 200:
        pass
    case 408:  # Request Timeout
        # We may want to retry
    case 429:  # Too Many Requests
        # We may want to retry later...
    case _:
        # We probably won't be able to recover from the error, raise an error?
```

### Raise a custom error

As API calls can raise errors, they need to be handled on the caller's side by catching them.

```python
# In my app
try:
    response = vespa_app.query(body)
except HTTPError as e:
    # Do we want to something for 408 or 429?
except VespaError as e:
    # Do we want to read the error message to decide how to handle the error?

# or we simply let the error be raised when it occurred.
response = vespa_app.query(body)
```

The former approach is more flexible, while the latter takes a more error-agnostic approach.

kkraune:
I merged #498 for release today. I supposed we can close this then?

rejasupotaro:
@kkraune Yes, thank you for your review 👍 🍺

kkraune:
Awesome! We appreciate the improvements in usability and error handling, keep the PRs coming :-)

PR #501, opened 2023-05-04
method_whitelist replaced by this MERGEOK

kkraune:
- lets give this a test run - seems like pyvespa fails with newer urllib3.util versions


kkraune:
looks like the right change, @lesters , good to review and merge

lesters (approved):

PR #510, opened 2023-05-30
Support setting default stemming of Schema

maxice8:
This adds support for setting the default stemming of a schema by adding
a 'stemming: <stemming-type>' to the root of the schema, as described in https://docs.vespa.ai/en/reference/schema-reference.html#schema

### DIFF
--- a/vespa/package.py
+++ b/vespa/package.py
@@ -1098,6 +1098,7 @@ class Schema(object):
         global_document: bool = False,
         imported_fields: Optional[List[ImportedField]] = None,
         document_summaries: Optional[List[DocumentSummary]] = None,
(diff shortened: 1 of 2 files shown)


kkraune:
Thanks! Method signatures are growing large. Maybe we should consider using kwargs?

maxice8:
I'll switch them to use kwargs as described and implemented the [first time we ran into this](https://github.com/vespa-engine/pyvespa/pull/414), Should I convert all of them or just the new ones?

maxice8:
> I would say just new ones to not break things now. We can clean this up laster at some major version update.

Should be good to go then

lesters (approved):
LGTM!

PR #511, opened 2023-06-05
Improvements and Fixes to Validation-Overrides

maxice8:
Commits contain full explanations:

1. Fix bug with missing validation-overrides.xml when using to_files()
2. Don't make :comment: an empty string and then render as comment="", just keep as None and don't try rendering it
3. Add an Enum that contains all known Validation IDs and check Validation against it, keep accepting a string for backwards compat and custom vespas


lesters (approved):
LGTM!

Issue #343, opened 2022-06-03
Remove data plane certificate and key management from pyvespa


thigm85:
pyvespa currently generates a data plane certificate and key and stores it in a file every time a deployment is made. This behavior conflicts with the workflow of using the vespa-cli to generate API key and dataplane certificate and key.

I suggest we remove this functionality from pyvespa to rely solely on vespa-cli to set-up certificates and keys for Vespa Cloud interaction.

thomasht86:
Closed by #809 
This is now done automatically by the `vespacli`-module

PR #513, opened 2023-06-29
Use cert/key generated by 'vespa auth cert' instead of regenerating each deployment

tmaregge:
Pyvespa used to make a new cert/key pair each time an app was deployed.
This pair was different than the one made by 'vespa auth cert', which meant that you could not use CLI commands to interact with an app deployed with pyvespa.
The PR makes pyvespa look for an existing pair in either the local .vespa directory (in application root), or in ~/.vespa.
If it can't find any, the deployment fails.

### DIFF
--- a/vespa/deployment.py
+++ b/vespa/deployment.py
@@ -378,8 +378,8 @@ class VespaCloud(object):
                 serialization.PublicFormat.SubjectPublicKeyInfo,
             )
         )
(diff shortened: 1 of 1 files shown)


tmaregge:
Related to this issue: https://github.com/vespa-engine/pyvespa/issues/343

lesters (approved):
Great!

PR #514, opened 2023-07-03
Tmaregge/cloud deploy from disk

tmaregge:
Commit 1 implements a method to deploy to Vespa Cloud from an application package located on disk. This was already supported for Docker deployments, but not for cloud.

Commit 4 makes a base class for shared functionality between VespaDocker and VespaCloud to reduce duplication.

Commit 5 rearranges the order of the methods in deployment.py to improve readability without changing the functionality of the program. Previously, public and private methods were declared here and there, seemingly randomly. Now, the order is: magic methods, public methods, private methods.

### DIFF
--- a/vespa/application.py
+++ b/vespa/application.py
@@ -218,10 +218,6 @@ class Vespa(object):
         """
         endpoint = "{}/ApplicationStatus".format(self.end_point)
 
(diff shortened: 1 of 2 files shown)


lesters (approved):
LGTM!

PR #515, opened 2023-07-04
Reintroduce pyvespa-generated certs as a fallback if other certs are missing

tmaregge:
In order not to break deployments for users who have not yet generated their cert/key with 'vespa auth cert', the old way of generating pyvespa certs has been readded as as a fallback option, in case certs are not found elsewhere.
A warning is printed recommending the user to switch to using the CLI for certs.


lesters (approved):
LGTM!

hmusum (approved):
👍

PR #547, opened 2023-08-22
Lesters/add non document field

lesters:
@kkraune Please review.


kkraune (approved):
using macro made this much better, too!

PR #553, opened 2023-08-28
Better handling of timeout and resets and increase timeout to align with vespa

lesters:
@kkraune Please review.
@jobergum FYI

Ref: https://github.com/vespa-engine/vespa/issues/27648


kkraune (approved):
👍

PR #558, opened 2023-09-04
Add nearest neighbor example

kkraune:
... or @lesters


jobergum:
I think this example should be focused on what 99% are using vector search for, with angular distance.

PR #573, opened 2023-10-05
Add global-phase support and data plane token support

jobergum:
- Add global-phase support
- Add option for using token based authentication for data plane
- Simplify handling of session parameters

What is left:
- Integration tests need to be updated to test end to end.
- Add ability to read multiple endpoints for vespa cloud (mtls + token)


kkraune (dismissed):
💯

vespa/application.py
```diff
@@ -105,18 +106,20 @@ def __init__(
...
+        vespa_cloud_secret_token: Optional[str] = None,
         output_file: IO = sys.stdout,
         application_package: Optional[ApplicationPackage] = None,
     ) -> None:
         """
         Establish a connection with an existing Vespa application.
 
-        :param url: Vespa instance URL.
-        :param port: Vespa instance port.
+        :param url: Vespa data plane endpoint.
```
lesters, line 116: `data plane` is a bit confusing for users that want to connect to Vespa instances not on our cloud, for instance a locally running Docker instance?

lesters: This is just nit picking, but perhaps we should leave it as `endpoint` and `port` - and mention data planes etc in the param description below?

lesters (dismissed):
Great!

vespa/application.py
jobergum: Agree

jobergum: Fixed with https://github.com/vespa-engine/pyvespa/pull/573/commits/02ccd75b18f68950496e32ce897daf6b26e55854

kkraune (approved):
Awesome! 💯 x 💯

PR #587, opened 2023-10-16
remove reply-after and some re-factoring with todos

jobergum:

### DIFF
--- a/vespa/application.py
+++ b/vespa/application.py
@@ -32,6 +32,7 @@ retry_strategy = Retry(
     allowed_methods=["POST", "GET", "DELETE", "PUT"],
 )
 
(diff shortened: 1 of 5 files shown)


vespa/application.py
```diff
@@ -996,7 +999,7 @@ def get_data(
             self.app.end_point, namespace, schema, str(data_id)
         )
         response = self.http_session.get(end_point)
-        raise_for_status(response)
+        raise_for_status(response) # This is qustionable, since it will throw on 404 as well.
```
kkraune, line 1002:
```suggestion
        raise_for_status(response) # This is questionable, since it will throw on 404 as well. 
```

kkraune (approved):

kkraune (approved):

PR #599, opened 2023-10-25
Major update

jobergum:
- Deprecate batch APIs
- Do not throw on retry errors without capturing actual error. Previously any retry error would just throw and making it difficult to assess the actual reason such as out of storage, issues in the feed or a query timeout. 
- More unit tests 
- Saner default connections for Vespa:query operations where you only have a single request and no reason for setting up a large connection pool
- `*kwargs` for query and document v1 API to send valid parameters directly 
- fix bug limiting the throughput of feed_iterable. 

### DIFF
--- a/docs/sphinx/source/query.ipynb
+++ b/docs/sphinx/source/query.ipynb
@@ -55,6 +55,7 @@
    "source": [
     "from vespa.application import Vespa\n",
     "from vespa.io import VespaQueryResponse\n",
+    "from vespa.exceptions import VespaError\n",
     "\n",
     "app = Vespa(url=\"https://api.cord19.vespa.ai\")"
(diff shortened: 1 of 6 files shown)


kkraune:
Supert! Det vi skriver "No replacement is planned" kan vi kanskje heller bare linke til https://pyvespa.readthedocs.io/en/latest/reads-writes.html#Batch og så skriver vi noe saklig der om bruk av det nye funksjonene for å stream inn?

kkraune (dismissed):

jobergum:
>No replacement is planned"

Will update the doc and notebooks etc after this PR goes through pipelines.

lesters (approved):
Great!

PR #629, opened 2023-11-10
Jobergum/streaming mode

jobergum:

### DIFF
--- a/vespa/application.py
+++ b/vespa/application.py
@@ -2,6 +2,7 @@
 
 import sys
 import ssl
+from tokenize import group
 import aiohttp
 import asyncio
 import requests
@@ -264,7 +265,7 @@ class Vespa(object):
 
     def query(
         self,
-        body: Optional[Dict] = None, **kwargs
+        body: Optional[Dict] = None, groupname:str=None, **kwargs
     ) -> VespaQueryResponse:
         """
         Send a query request to the Vespa application.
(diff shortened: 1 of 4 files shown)


vespa/templates/services.xml
```diff
@@ -11,6 +11,9 @@
         {% if schemas %}
         <search></search>
         <document-api></document-api>
+        <document-processing></document-processing>
+        {% endif %}
+        {% if search_clusters %}
         {% endif %}
```
jonmv, line 17: Isn't this just empty?

vespa/application.py
```diff
@@ -979,9 +1005,10 @@ async def update_data(
         fields: Dict,
         create: bool = False,
         namespace: str = None,
+        groupname:str=None,
         **kwargs
     ) -> VespaResponse:
-        path = self.app.get_document_v1_path(id=data_id, schema=schema, namespace=namespace)
+        path = self.app.get_document_v1_path(id=data_id, schema=schema, namespace=namespace, group=groupname)
```
jonmv, line 1011: Just string groups? I guess that's the most common use-case.

jonmv (approved):

vespa/application.py
jobergum: According to doc, group is the way https://docs.vespa.ai/en/streaming-search.html#using-streaming-search

jonmv (approved):

vespa/application.py
jonmv: 👍

PR #631, opened 2023-11-16
Fix typo

kkraune:


kkraune:
work note:
```
find . -name \*.py -or -name \*.ipynb | xargs sed -i.orig "s/is_successfull()/is_successful()/g"
find . -name \*.orig | xargs rm
```

jobergum:
For this we also need to change vespa.io.VespaResponse.is_successfull - I suggest that we unbreak this by adding a new method and deprecate the misspelled variant?

kkraune:
right, will update PR

kkraune:
the regexp did of course not catch `is_successfull(self)` - adding now

jobergum (approved):

PR #633, opened 2023-11-23
Recognize 'alias' as an option for Field

maxice8:


maxice8:
Updated it to treat alias as an `List[str]` instead of just `str` allowing multiple aliases to be added to a Field.

maxice8:
Sure, I have added a test to it but I am not sure it is what you wanted (we didn't have those test requirements back when I was more actively contributing), please take a look and let me know if I made the tests correctly and what I need to change if not.

jobergum (approved):

jobergum:
Thank you very much for your contribution! Yes, generally great to have unit test coverage

PR #634, opened 2023-11-23
modernize

maxice8:
While I was implementing #633 and reading on feeding_iterable to move over from feed_batch I ran into these. 

- Hint that __repr__ dunders return a str
- Remove duplicate import of 'traceback' module
- Annotate parameters of __eq__, its return, and return NotImplemented


jobergum (approved):
Thank you!

PR #653, opened 2023-12-29
Call the callback function without using keyword argument

rejasupotaro:
## Overview

This PR allows defining the callback function of `feed_iterable` with arbitrary arg names.

## Background

I recently updated pyvespa to 0.39.0 and replaced `feed_batch` with the newly introduced function `feed_iterable`. I called `feed_iterable` for testing and got the error `TypeError: callback() got an unexpected keyword argument 'id'`.

Looking into the implementation, I found that `_handle_result_callback` invokes the callback with `id=id`, which was the cause of my issue.

Screenshot 2023-12-29 at 22 38 35

_☝️ How it occurs_

In Python, `id` is a built-in function ([Built-in Functions — Python 3.12.1 documentation](https://docs.python.org/3/library/functions.html#id)), so I usually refrain from using it as a variable name. I guess others might encounter this error as well

## Solution

I stopped calling the argument of that callback function as a keyword argument. It doesn't change the behavior but would give more flexibility.

Screenshot 2023-12-29 at 22 40 12

---

### DIFF
--- a/vespa/application.py
+++ b/vespa/application.py
@@ -306,15 +306,15 @@ class Vespa(object):
                 schema=schema, data_id=data_id, fields=fields, namespace=namespace, groupname=groupname, **kwargs
             )
    
-    def feed_iterable(self, 
-        iter:Iterable[Dict], 
-        schema:Optional[str] = None, 
-        namespace:Optional[str] = None, 
-        callback:Optional[Callable] = None, 
-        operation_type:Optional[str] = "feed",
-        max_queue_size:int = 1000,
-        max_workers:int = 8, 
-        max_connections:int = 16,
+    def feed_iterable(self,
+        iter: Iterable[Dict],
+        schema: Optional[str] = None,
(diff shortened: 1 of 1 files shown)


vespa/application.py
```diff
@@ -414,7 +414,7 @@ def _submit(doc:dict, sync_session:VespaSync) -> Tuple[str, Union[VespaResponse,
             except Exception as e:
                 return (id, e)
 
-        def _handle_result_callback(future:Future, callback:Callable):
+        def _handle_result_callback(future: Future, callback: Optional[Callable[[VespaResponse, str], None]]):
```
rejasupotaro, line 417: We can assume that the type of the callback function is `Callable[[VespaResponse, str], None]` as documented

```python
:param callback: A callback function to be called on each result. Signature `callback(response:VespaResponse, id:str)`
```

Also, it's optional.

jobergum (approved):
Great! Thanks a lot for your contribution 💯

PR #657, opened 2024-01-08
Exclude test files from the package

rejasupotaro:
## Overview

When writing code, I noticed that the `vespa` package has modules for its testing.

Screenshot 2024-01-09 at 7 48 46

Following the common practice, I moved test files to the `tests` directory not to include them in the package.

While this PR does not alter the tests themselves, it does modify the project's structure ⚠️ However, as the project management style is up to the owner, feel free to close this PR if this approach is not preferable.

## Files in Package

Python includes all files in a directory as part of the package. We can see what files are included by unzipping the built package.

<details>

<summary>Before</summary>

```
$ python -m build
$ cd dist
$ unzip pyvespa-0.7.0+dev-py3-none-any.whl
$ tree vespa
vespa
├── __init__.py
├── application.py
├── deployment.py
├── exceptions.py
├── io.py
├── package.py
├── py.typed
├── templates
│   ├── hosts.xml
│   ├── macros.txt
│   ├── query_profile.xml
│   ├── query_profile_type.xml
│   ├── schema.txt
│   ├── services.xml
│   └── validation-overrides.xml
├── test_application.py
├── test_integration_docker.py
├── test_integration_vespa_cloud.py
├── test_integration_vespa_cloud_token.py
├── test_integration_vespa_cloud_vector_search.py
├── test_io.py
└── test_package.py

1 directory, 21 files
```

</details>

If we have the `tests` directory, test files are now not included in the package as shown below.

<details>

<summary>After</summary>

```
$ python -m build
...
$ tree vespa
vespa
├── __init__.py
├── application.py
├── deployment.py
├── exceptions.py
├── io.py
├── package.py
├── py.typed
└── templates
    ├── hosts.xml
    ├── macros.txt
    ├── query_profile.xml
    ├── query_profile_type.xml
    ├── schema.txt
    ├── services.xml
    └── validation-overrides.xml
```

</details>

Other than this, it is convenient if we have a separate directory for tests when adding more tests using fixtures https://docs.pytest.org/en/6.2.x/fixture.html.

## Dependency Management

Apart from the main topic of this PR, I encountered the following error while writing tests for another PR.

```
(.venv) ~/p/pyvespa ❯❯❯ pytest
=========================================================================================================== test session starts ============================================================================================================
platform darwin -- Python 3.8.15, pytest-7.2.2, pluggy-1.0.0
rootdir: /Users/kentaro-takiguchi/projects/pyvespa
collected 74 items / 1 error

[...]

PR #660, opened 2024-01-16
Minor fix for dependency management

rejasupotaro:
## Overview

This PR fixes a minor issue in dependency management. More specifically, it replaces `.[full]` with `.[dev]`.

I found this when I was looking into https://github.com/vespa-engine/pyvespa/pull/659

## Context

`pip install -e '.[dev]'` was introduced in https://github.com/vespa-engine/pyvespa/pull/657 but I overlooked the use of `.[full]` in `requirements.txt`.

However, it wouldn't cause an immediate problem for the following reasons:

- Executing `pip install -e '.[full]'` does not raise an error even if `.[full]` is not defined
- In the current settings, dependencies are always installed in the following order

```
# from screwdriver.yaml
python3 -m pip install --upgrade pip
python3 -m pip install -e .[dev]
python3 -m pip install -r docs/sphinx/source/requirements.txt
python3 -m pip install -r docs/sphinx/source/notebook_requirements.txt
```

But I would like to fix it to prevent future confusion.

---

### DIFF
--- a/docs/sphinx/source/requirements.txt
+++ b/docs/sphinx/source/requirements.txt
@@ -1,4 +1,4 @@
-.[full]
+.[dev]
 sphinx
 ipykernel
 nbsphinx


jobergum (approved):

jobergum:
Fantastic! Thanks! We don't pin llama-index,  because we want to know when they break the API so that we can fix the notebook. They are moving fast.

jobergum:
All good now - thanks again @rejasupotaro for your time on this!

image

rejasupotaro:
Great news to start my day 👏 Thank you for your help!

PR #664, opened 2024-01-19
streaming search example with colbert ranking

andreer:


jobergum:
You need to rename the notebook to have -cloud as suffix as the rest of the notebooks, that is why the Screwdriver/7055/PR:notebooks-except-cloud fails

jobergum (approved):

PR #671, opened 2024-02-04
eval matryoshka embeddings

andreer:


jobergum (dismissed):
Looks great - since the whole notebook is one one large line: 

- Remove quotes "[Matryoshka Representation Learning](https://arxiv.org/abs/2205.13147)". Also consider rewriting with less fluff words (very, elegantly, easily). 
-  Try to summarize the experiment better in the intro. It doesn't appear to use tensor slicing but a phased ranking pipeline as described in the original MRL paper. Then the key differentiator is that Vespa allows phased ranking and multiple vector fields and where the second phase can use a paged version from disk. 
- You are always asking for the largest versions and then chopping the vector in the python code.  It would be nice to see an experiment with using the custom chopped version for ingestion but ask openai for a shortened version for for the experiments with the reduced dimensionality without re-ranking as this would hopefully reduce the time to obtain the embedding (less data over the network).

- In the case of re-ranking you will need the larger version, and chopping is great because you don't need two calls. 
- Only define the input tensors needed for the rank profile 
- Add an explanation of that the overall measure is the average nDCG@10 score for all the 50 queries - and that the notebook trims 100
- Must look at the baseline as it seems low

jobergum (approved):
Great!

PR #683, opened 2024-02-15
try legacy mode then

jobergum:


jobergum:
Trying the legacy route for now until we can get some guidance on how the new API is supposed to be used https://github.com/run-llama/llama_index/issues/10747

bjormel (approved):

PR #685, opened 2024-02-21
Update application package name checks

tmaregge:
This PR adds more checks to make sure the application package name is valid in Vespa.

Previously, some invalid app package names could be used and it would not be noticed until deployment, where it would fail with a content hash mismatch (e.g. [https://github.com/vespa-engine/vespa/issues/30219](https://github.com/vespa-engine/vespa/issues/30219)), which was very confusing.

### DIFF
--- a/vespa/package.py
+++ b/vespa/package.py
@@ -2076,9 +2076,9 @@ class ApplicationPackage(object):
         It will create a default :class:`Schema`, :class:`QueryProfile` and :class:`QueryProfileType` that you can then
         populate with specifics of your application.
         """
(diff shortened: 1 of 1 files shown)


tmaregge:
Note: These tests are a bit strict. Vespa actually supports dashes in application package names (but not double dashes). Pyvespa could probably also support this, but it would require more elaborate tests.

jobergum (dismissed):

jobergum:
Consider adding a test for this.

jobergum (approved):

PR #725, opened 2024-04-04
Add monitoring to prod deployments

tmaregge:
This PR adds monitoring to production deployments via Pyvespa, similar to how it's done for dev deployments. This way, you don't *have* to go to the web console to figure out what's going on. Also needed for integration test.


kkraune (approved):
This is WIP

PR #730, opened 2024-04-08
Add `feed_iterable_async()`-method

thomasht86:

This PR got a bit bigger than intended. Some changes are due to linting. 
The main new feature of this PR is the `feed_iterable_async()`-method, along with a preliminary notebooks, which demonstrates the use of it and runs a simplified benchmark.  

image

While adding this I also grabbed some opportunities for refactoring. 

Both unit and integration-tests are added for this method. ✅

Would love some input from @jobergum on improvements for notebook. 
Should I for example test feeding against Cloud, or maybe do a more scientific approach by feeding documents of a specific size (and amount of docs?)


kkraune (dismissed):

jobergum (changes requested):
Generally, this is way too big for a single PR.

Don't bundle many things in the same PR. This PR refactors a lot of files, adds dependencies and introduces new functionality (+7500, -614).  It becomes unmanageable to review all these changes. 


- The feeding performance notebook does not compare apples with apples, upserts with create true requires a random read, comparing that with simple put which is just append only. Use at least 100K operations 

- The new async feed_iterable must support streaming inputs and not consume the iterator in a single go like the current implementation as you will run out of memory on the client when consuming non-trivial amounts of data.

PR #732, opened 2024-04-11
Adding feeding performance notebook

thomasht86:

Took a step back and adopted a more scientific approach trying to compare apples to apples, and using the `VespaAsync` as it already is implemented, resisting the urge to change the library code (for now) 😆 

Made it more adaptable, by defining parameters to be explored at the top

```python
num_docs = [1000, 10_000, 100_000]

params_by_function = {
    "feed_sync": {
        "num_docs": num_docs,
        "max_connections": [16, 64],
    },
    "feed_async": {
        "num_docs": num_docs,
        "max_connections": [16, 64],
        "num_concurrent_requests": [1000, 10_000],
    },
    "feed_iterable": {
        "num_docs": num_docs,
        "max_connections": [64, 128],
        "max_workers": [16, 64],
        "max_queue_size": [1000, 10000],
    },
}
```
feed_performance

- Had to reduce num_docs to only 1000 as running the notebook made Screwdriver time out. Added the larger experiment plots as static images to the notebook instead.


kkraune (approved):
Great! I did not go through the notebook, the approach look great!

jobergum (dismissed):
Great! Consider the following optional changes 

- If we want to promote pyvespacli package we should put it under our wings, it's currently a personal project 
- Remove the SSL warning
- Explain why we use a dummy schema with summary only and that as we add more features, we expect that the Vespa backend becomes the bottleneck

-  It seems like an unfair comparison to me as the vespa cli works with a single json file already consumed from S3, while the pyvespa parts are operating over a streamed response from S3? In that respect it might be that the conclusion is a bit off, in my experience there is not 2 orders of magnitude difference
- Consider writing more about why the synch _feed one and one operation is very latency sensitive (100 ms latency end to end => 10 feed operations per sec), 50 ms, =>20 etc.

thomasht86:
> Great! Consider the following optional changes
> 
>      If we want to promote pyvespacli package we should put it under our wings, it's currently a personal project
I am talking to Frode about how we can do transfer ownership etc. Not promoting it yet. 
> 
>     * Remove the SSL warning
✅ 
> 
>     * Explain why we use a dummy schema with summary only and that as we add more features, we expect that the Vespa backend becomes the bottleneck
✅  
>     * It seems like an unfair comparison to me as the vespa cli works with a single json file already consumed from S3, while the pyvespa parts are operating over a streamed response from S3? In that respect it might be that the conclusion is a bit off, in my experience there is not 2 orders of magnitude difference
Yeah, that´s true. Added a note about this, and pointed out it is an unfair comparison.  
>     * Consider writing more about why the synch _feed one and one operation is very latency sensitive (100 ms latency end to end => 10 feed operations per sec), 50 ms, =>20 etc.
Added some text with a practical example like this. 





> Great! Consider the following optional changes
> 
>     * If we want to promote pyvespacli package we should put it under our wings, it's currently a personal project
> 
>     * Remove the SSL warning
> 
>     * Explain why we use a dummy schema with summary only and that as we add more features, we expect that the Vespa backend becomes the bottleneck
> 
>     * It seems like an unfair comparison to me as the vespa cli works with a single json file already consumed from S3, while the pyvespa parts are operating over a streamed response from S3? In that respect it might be that the conclusion is a bit off, in my experience there is not 2 orders of magnitude difference
> 
>     * Consider writing more about why the synch _feed one and one operation is very latency sensitive (100 ms latency end to end => 10 feed operations per sec), 50 ms, =>20 etc.

thomasht86:
Need new approval after implementing feedback @jobergum @kkraune 🙏

kkraune (approved):

jobergum:
Great! 💯

Issue #718, opened 2024-04-02
update only supports assign updates


jobergum:
The update takes the input and just assumes everything is assign updates, ignoring other types of supported partial updates. 

```vespa_format = {"fields": {k: {"assign": v} for k, v in fields.items()}}```

Full context 
```
def update_data(
        self,
        schema: str,
        data_id: str,
        fields: Dict,
        create: bool = False,
        namespace: str = None,
        groupname:str = None,
        **kwargs
    ) -> VespaResponse:
        """
        Update a data point in a Vespa app.

        :param schema: The schema that we are updating data.
        :param data_id: Unique id associated with this data point.
        :param fields: Dict containing all the fields you want to update.
        :param create: If true, updates to non-existent documents will create an empty document to update
        :param namespace: The namespace that we are updating data.
        :param groupname: The groupname used to update data
        :param kwargs: Additional HTTP request parameters (https://docs.vespa.ai/en/reference/document-v1-api-reference.html#request-parameters)
        :return: Response of the HTTP PUT request.
        :raises HTTPError: if one occurred
        """
        
        path = self.app.get_document_v1_path(id=data_id, schema=schema, namespace=namespace, group=groupname)
        end_point = "{}{}?create={}".format(
            self.app.end_point, path, str(create).lower()
        )
        vespa_format = {"fields": {k: {"assign": v} for k, v in fields.items()}}
        response = self.http_session.put(end_point, json=vespa_format, params=kwargs)
        raise_for_status(response)
        return VespaResponse(
            json=response.json(),
            status_code=response.status_code,
            url=str(response.url),
            operation_type="update",
        )
```

PR #738, opened 2024-04-16
Allow other partial updates than `assign`

thomasht86:

Closing #718 

**What`s done?**

- Added `auto_assign`-parameter to `update_data()`-methods for both `VespaSync` and `VespaAsync`.
This is set to `True` by default, so the default behavior will remain the same. 
If `auto_assign` is set to `False`, however, the user can put any update request they want, including arithmetic, tensor operations etc. 
- Added some tests just for demonstration, but considered it infeasible to add tests for all update operations. Also, if request is invalid, the server will inform about this. 

Time spent on the feature: 10%
Time spent on making the tests pass: 90%. 
Hope (and think) I learned something from it.

### DIFF
--- a/tests/integration/test_integration_docker.py
+++ b/tests/integration/test_integration_docker.py
@@ -266,6 +266,7 @@ class TestApplicationCommon(unittest.TestCase):
         field_to_update,
         expected_fields_from_get_operation,
         expected_fields_after_update: Optional[Dict] = None,
+        **kwargs,
     ):
         """
         Feed, get, update and delete data to/from the application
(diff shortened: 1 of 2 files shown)


jobergum (dismissed):

jobergum:
Looks great, have a look at the test run failure?

kkraune (approved):

thomasht86:
> Looks great, have a look at the test run failure?

Test fixed. ✅ 
Note to self: Make sure tests passes *before* I ask for review next time

jobergum (approved):

jobergum:
Great!

PR #741, opened 2024-04-17
simplify, strip macros in all files

kkraune:


thomasht86 (approved):
Need some explanation to understand what this does, but it did seem to do the trick 🙌

Issue #734, opened 2024-04-11
document v1  paths should be urlencoded


jobergum:
```
{ "id": "doc#paragraph",  "fields": {..}}
```

Will likely cause a POST with #paragraph  in the id part and where # is a reserved char so you end up indexing just "doc" as the id :/

PR #743, opened 2024-04-19
734 document v1 paths should be urlencoded

thomasht86:

Closing #734.

Made the function that creates the path do quoting and changed unit test accordingly.


kkraune (approved):

kkraune:
This _might_ be a problem for some existing users, but I think we should just mention it in the release notes, as it is the right thing to do

PR #744, opened 2024-04-19
Add notebook to demonstrate use of `Snowflake/snowflake-arctic-embed-s`

thomasht86:

This notebooks demonstrates the use of these new embedding models: 

They require some additional parameters compared to other embedders we have demo'ed: 

```xml
 <component id="snow" type="hugging-face-embedder">
      <transformer-model url="https://huggingface.co/Snowflake/snowflake-arctic-embed-l/resolve/main/onnx/model_int8.onnx"/>
      <tokenizer-model url="https://huggingface.co/Snowflake/snowflake-arctic-embed-l/raw/main/tokenizer.json"/>
      <normalize>true</normalize>
      <pooling-strategy>cls</pooling-strategy>
    </component>
 ```

This is not self-explanatory in pyvespa 😅 

```python
                Parameter(
                    "normalize",
                    {},
                    "true",
                ),
                Parameter(
                    "pooling-strategy",
                    {},
                    "cls",
                ),
```
 
 It is also really nice to see (and show) that most new embedding models now seem to get ONNX-model artifacts automatically on huggingface, so that Vespa users do not have to download models and zip in package, which often leads to issues if they include several large models etc. 
 
 As always, feedback makes me better, and is very welcome.


docs/sphinx/source/examples/evaluating-with-snowflake-arctic-embed.ipynb
```diff
@@ -0,0 +1,1774 @@
...
+    "<picture>\n",
+    "  <source media=\"(prefers-color-scheme: dark)\" srcset=\"https://vespa.ai/assets/vespa-ai-logo-heather.svg\">\n",
+    "  <source media=\"(prefers-color-scheme: light)\" srcset=\"https://vespa.ai/assets/vespa-ai-logo-rock.svg\">\n",
+    "  <img alt=\"#Vespa\" width=\"200\" src=\"https://vespa.ai/assets/vespa-ai-logo-rock.svg\" style=\"margin-bottom: 25px;\">\n",
+    "</picture>\n",
+    "\n",
+    "# Evaluating retrieval with Snowflake arctic embed\n",
+    "\n",
+    "This notebook will demonstrate how different rank profiles in Vespa can be set up and evaluated.\n",
+    "For the rank profiles that use semantic search, we will use the small version of [Snowflake`s arctic embed model series](https://huggingface.co/Snowflake/snowflake-arctic-embed-s) for generating embeddings.\n",
```
kkraune, line 16: fix link parsing (maybe)
```suggestion
    "For the rank profiles that use semantic search, we will use the small version of [Snowflake's arctic embed model series](https://huggingface.co/Snowflake/snowflake-arctic-embed-s) for generating embeddings.\n",
```

kkraune (dismissed):
The workbook worked great for me! I did not dig into the details, super with the NDCG eval/plot :-)

docs/sphinx/source/examples/evaluating-with-snowflake-arctic-embed.ipynb
thomasht86: Sorry, still have some way to go with the Mac-keyboard :D

jobergum:
Great! The PR says small, but does the code use large? 

- We need to guard against log(0) because it will cause a NaN/-Infinity when there is no keyword matches. I think that this is demonstrated in a sample app, I'll try to find it and correct this. 
- Consider using a library nDCG implementation unless you are 100% confident that the implementation is correct. Maybe there is an educational aspect here

jobergum (approved):

thomasht86:
Great input. It is definitely an education aspect for me. 
Maybe others with my experience level will appreciate it as well, was my thought. 
I will do verification with scikit-learn implementation before publishing, but it is nice to avoid that dependency.
Also align reference to small/large.

PR #745, opened 2024-04-19
Make integration test for production deployments

tmaregge:
The PR adds an integration test that sets up a prod deployment, tests feeding and querying and then tears down the application. The test builds upon the existing vector search test. Deletion is handled by deploying with an empty deployment.xml. 

The PR also fixes what appears to be a bug in the way validation overrides were handled in pyvespa. Previously, validation overrides were not actually included in the app package zip.

### DIFF
--- a/tests/integration/test_integration_vespa_cloud.py
+++ b/tests/integration/test_integration_vespa_cloud.py
@@ -7,22 +7,17 @@ import pytest
 from requests import HTTPError
 import unittest
 from cryptography.hazmat.primitives import serialization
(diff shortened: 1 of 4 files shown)


tests/integration/test_integration_vespa_cloud.py
```diff
@@ -13,6 +13,7 @@
     TestApplicationCommon,
     create_msmarco_application_package,
 )
+from pathlib import Path
```
thomasht86, line 16:
```suggestion
```
Unused import. Probably used during testing? :)

thomasht86 (approved):
This is great! 💯 
Let´s remember to update the docs related to prod deployment on next release.

PR #751, opened 2024-04-23
Add data plane auth token to integration test

tmaregge:
One of the vector search integration tests fails on the CI server. Adding the same auth_client_token_id as for the other test should hopefully fix this.


thomasht86 (approved):
Good!
Let's see if it works now 🫰

Issue #746, opened 2024-04-19
Replace old logos with new in notebooks


thomasht86:
See for example https://pyvespa.readthedocs.io/en/latest/getting-started-pyvespa-cloud.html

PR #753, opened 2024-04-25
Replace old logo

thomasht86:

Closing #746 

I thought that the old logo was present several places, but this was the only one I could find.


thomasht86:
The image is the only thing that is changed, the remaining diff is due to ruff formatting. 
(We haven't yet batch formatted untouched files.)

kkraune (approved):

Issue #748, opened 2024-04-22
Async data operations does not apply retry strategy


thomasht86:
https://github.com/vespa-engine/pyvespa/blob/90d5342eefe9368cf6527fa7a042630e49c179ab/vespa/application.py#L1099C1-L1122C10

Tenacity's default policy is to `retry_if_exception_type`. 
https://tenacity.readthedocs.io/en/latest/
A 4/5xx-response will in this case never raise an exception, and thus not cause retry.

PR #756, opened 2024-04-30
Jonmv/retry 429 and 503 when feeding

jonmv:
@thomasht86 please review.

### DIFF
--- a/vespa/application.py
+++ b/vespa/application.py
@@ -66,6 +66,16 @@ def raise_for_status(
             raise VespaError(error_message) from http_error
         raise HTTPError(http_error) from http_error
 
+def retry_docv1(state):
+    return state.outcome.failed or state.outcome.result().status >= 400
+
+def stop_docv1(state):
+    if state.outcome.failed or state.outcome.result().status == 503: return state.attempt_number >= 3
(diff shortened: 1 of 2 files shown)


jonmv:
Urgh, this turned ugly, but it seems to be the only way to make it work, perhaps because of the typing—when testing a similar setup locally, the simpler form worked perfectly.

thomasht86 (approved):
Great. I don`t think it got _that_ ugly. 👍

thomasht86:
Fixes #748

vespa/application.py
```diff
@@ -1096,7 +1110,19 @@ async def query(
             json=await r.json(), status_code=r.status, url=str(r.url)
         )
 
-    @retry(wait=wait_exponential(multiplier=1), stop=stop_after_attempt(3))
+    @retry(
+        wait=wait_exponential(multiplier=1),
+        retry=retry_any(
+            retry_if_exception(lambda x: True),
```
jonmv, line 1116: Could consider being more selective here, later, but I don't know what errors/exceptions to expect from the HTTP library, and I couldn't find documentation of that either. We want to retry transport-related failures, but not SSL-related, and other user errors.

PR #761, opened 2024-05-06
Add validation override for cert removal

tmaregge:
The prod deployment integration test appears to be failing because the certificate is overwritten. This is the default behavior in pyvespa, and works fine in dev, but prod is more strict. The PR adds a validation override to get around this (at least for now).


thomasht86 (approved):
Ok, but let's consider using CLI-generated cert going forward.

PR #772, opened 2024-05-13
Use jitter and max 10s backoff for async docv1 client

jonmv:
@thomasht86


thomasht86 (approved):
Great! 💯  
This will make sure all tasks are not retried at the same time, causing a wave of 429s/backoffs 😄

PR #782, opened 2024-05-22
Fix test (kernel died)

thomasht86:

2 changes:
1. Add `vespacli` to `notebook_requirements.txt`, so it gets pre-installed on test-runners, reducing execution time of each notebook.
2. Try to fix strange test failure while running https://pyvespa.readthedocs.io/en/latest/examples/Matryoshka_embeddings_in_Vespa-cloud.html by reducing memory usage.

Test fails with
```python
08:04:58 nbclient.exceptions.DeadKernelError: Kernel died 
```
on some of the cells right after Cloud deployment.

This error usually occurs due to exceeding memory resources. 
It is still strange that this would only occur now, with unchanged resource specs. 🤔


jobergum:
💯

PR #776, opened 2024-05-16
Expose visit api

olaughter:
Adds a method that returns a generator which fetches all data across the documents in a vespa instance. Optionally run concurrently by increasing the number of slices, and optionally supply a yql query to filter.

I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

https://github.com/vespa-engine/pyvespa/issues/711

@thomasht86

PR #786, opened 2024-05-27
Doc fixes and visit example

thomasht86:

Some minor changes after #776:
- Rename `chunk_size` to `wanted_document_count`, to avoid misunderstanding the parameter. 
- Expand visit-example in `reads_writes.ipynb`. 
- Add example of `increment`-partial update to the same notebook.

Also, added a few more usage examples to the most used API`s. 
With this merged, I think we can do a release.


kkraune:
we might want to just add wanted_document_count for now and deprecate chunk_size - and remove it after some time, to break less usages. otherwise great

thomasht86:
> we might want to just add wanted_document_count for now and deprecate chunk_size - and remove it after some time, to break less usages. otherwise great



> we might want to just add wanted_document_count for now and deprecate chunk_size - and remove it after some time, to break less usages. otherwise great

Ok, is this still necessary, as `chunk_size` has not been in any release yet?

kkraune:
oh, then is it not necessary :-)

kkraune (approved):

PR #790, opened 2024-05-29
CI: pyvespa release pipeline

thomasht86:

This PR introduces a semi-automated release flow:

1. Create a new release from [github.com/vespa-engine/pyvespa/releases/new](https://github.com/vespa-engine/pyvespa/releases/new).
2. Make sure to tag the release with the version number, e.g., `v0.41.0`.
3. This tag will trigger a github action that will publish the package to [PyPI](https://pypi.org/project/pyvespa/).
4. A PR will also be automatically created to update the affected files with the new version. This PR should be merged to keep the version updated in the repository.

This workflow can also be dispatched manually, but note that steps 3 and 4 will ONLY be triggered by a release.

I think this is a good solution for now. 
Having autorelease on every commit to master can be a future option, but then we should ideally have all CI-tests done on PR. Now, some screwdriver jobs are only done on commit.


kkraune:
so this will release daily mon-thu if there are any commits?

thomasht86:
That is for the vespacli package (if there is a newer version available). 
(The change related to those files in this PR was due to basing this branch off #789, which is now merged).

For pyvespa, I suggest the semi-automated flow described above.
Feedback/inputs are welcome.

kkraune:
I see, right, makes sense

kkraune (approved):

kkraune (approved):
👍

Issue #791, opened 2024-05-30
Refine retry-strategy (429-errors) for VespaSync


thomasht86:
The sync HTTP-client currently retries only 3 times for 429-errors. 
Should do this many more times, preferrably with exponential backoff and random jitter. 

For 503, the current strategy is OK.

PR #794, opened 2024-05-30
Thomasht86/sync retry strategy 429

thomasht86:

Closing #791 

After @jonmv fixed the async retry-strategy, we decided fixing the sync retry strategy for 429 was highest priority to gain confidence that the upcoming change in feed queue handling will not affect users negatively. 

After a nice tip from @jonmv on how to easily simulate slow feeding, it was also easy to add integration test for this. 
These tests would have failed with the old strategy, where 429s were only retried 3 times, and are very relevant as people add slow embedders etc. 

- Also a small refactor to get rid of `WORK_DIR` env variable.

Will create a new issue to add async feeding to the same integration-tests. 

Really don't like to break the DRY-principle by duplicating a lot of code for both docker and cloud tests. 
It would be very nice to refactor the Test classes properly, to make it easier to add both Cloud and Docker applications without duplicating code.

### DIFF
--- a/vespa/application.py
+++ b/vespa/application.py
@@ -532,7 +532,9 @@ class Vespa(object):
                     )
 
         with VespaSync(
-            app=self, pool_maxsize=max_connections, pool_connections=max_connections
+            app=self,
+            pool_maxsize=max_connections,
+            pool_connections=max_connections,
         ) as session:
             queue = Queue(maxsize=max_queue_size)
             with ThreadPoolExecutor(max_workers=max_workers) as executor:
@@ -826,7 +828,9 @@ class CustomHTTPAdapter(HTTPAdapter):
 
 
 class VespaSync(object):
-    def __init__(self, app: Vespa, pool_maxsize: int = 10, pool_connections=10) -> None:
+    def __init__(
+        self, app: Vespa, pool_maxsize: int = 10, pool_connections: int = 10
+    ) -> None:
         self.app = app
         if self.app.key:
             self.cert = (self.app.cert, self.app.key)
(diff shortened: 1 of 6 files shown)


vespa/application.py
```diff
@@ -802,6 +796,35 @@ def get_document_v1_path(
...
+        )
+        self.retry_strategy = retry_strategy
+
+    def send(self, request, **kwargs) -> Response:
+        for attempt in range(self.num_retries_429):
+            response = super().send(request, **kwargs)
+            if response.status_code == 429:
+                wait_time = (
+                    2** attempt + random.uniform(0, 1)
+                )  # Exponential backoff with jitter. The 10th retry will sleep for 1024 seconds.
```
jonmv, line 821: I think that is a bit excessive, perhaps cap it at 10s or so, and even start a bit lower?

jonmv (dismissed):
Nice! I think we should probably avoid sleeping for more than 10s, as the user may just think the feeder is hanging.

vespa/application.py
jonmv: Did we decide to reduce this a bit, though? I can't remember now 😁

vespa/application.py
```diff
@@ -802,8 +798,39 @@ def get_document_v1_path(
...
+            allowed_methods=["POST", "GET", "DELETE", "PUT"],
+        )
+        self.retry_strategy = retry_strategy
+
+    def send(self, request, **kwargs) -> Response:
+        for attempt in range(self.num_retries_429):
+            response = super().send(request, **kwargs)
+            if response.status_code == 429:
+                wait_time = (
+                    2** attempt + random.uniform(0, 1)
```
jonmv, line 822:
```suggestion
                    0.1 * (1.618** attempt) + random.uniform(0, 1)
```

jonmv (approved):
Great!

vespa/application.py
thomasht86: Had a commit, but got distracted from pushing it. 
Great that you saw it! 👍

jonmv (approved):

PR #809, opened 2024-06-21
Prod deployment and interactive auth

thomasht86:

With this PR, prod deployment should work both from application package and from disk_folder. 
Removed functionality to track and wait for prod deployment, and implemented `check_production_build_status()`-method. 
Now, we no longer wait for a prod deployment, but return the build_no, which can be used to poll for status. 

Also expanded with two integration tests:
1. `TestMsmarcoProdApplicationWithTokenAuth`
2. `TestProdDeploymentFromDisk`

Also started on unit-tests for `deployment.py`, which was missing.


vespa/deployment.py
```diff
@@ -440,65 +452,149 @@ def __init__(
...
+        :param application_package: ApplicationPackage to be deployed. Either this or disk_folder must be set.
         :param key_location: Location of the control plane key used for signing HTTP requests to the Vespa Cloud.
         :param key_content: Content of the control plane key used for signing HTTP requests to the Vespa Cloud. Use only when
             key file is not available.
         :param auth_client_token_id: Use token based data plane authentication. This is the token name configured in the Vespa Cloud Console.
-            This is used to configure Vespa services.xml. The token is given read and write permissions.
+            This is used to configure Vespa services.xml. The token is given read and write permissions. If initiliazing from disk_folder, make sure
+            that services.xml is configured to use the provided token_id.
         :param output_file: Output file to write output messages. Default is sys.stdout
+        :param disk_folder: Disk folder for application root.
```
jonmv, line 508: I think `application_path` or `application_package_path` would be a better name. It's more in line with what we use already.

vespa/deployment.py
```diff
@@ -440,65 +452,149 @@ def __init__(
...
+                    ),
+                    AuthClient(
+                        id="token",
+                        permissions=["read,write"],
+                        parameters=[Parameter("token", {"id": auth_client_token_id})],
+                    ),
+                ]
+            else:
+                print(
+                    "Warning: Auth client token id set, but no application package provided. Make sure that services.xml is configured to use the provided token_id.",
```
jonmv, line 568: Is this really a warning, though? Is it not intended that users have a pre-existing package, and token auth? Certainly worth mentioning, but perhaps at a lower "level"?

vespa/deployment.py
```diff
@@ -550,127 +646,185 @@ def deploy(
...
 
         :param instance: Name of this instance of the application, in the Vespa Cloud.
-        :param disk_folder: Disk folder to save the required Vespa config files. Default to application name
-            folder within user's current working directory.
-        :param max_wait: Seconds to wait for the deployment.
+        :param disk_folder: Disk folder to either save the required Vespa config files (if initialized with application_package) or read them from (if initialized with disk_folder).
+        :param submit_options: Optional dictionary of submit options, containing one or more of the following keys:
+            - "repository"
+            - "branch"
+            - "commit"
```
jonmv, line 664: These are somewhat obsolete with `sourceUrl`, which is otherwise generated from these three on the server side. I think it's sufficient to just expose that one parameter.

vespa/deployment.py
```diff
@@ -550,127 +646,185 @@ def deploy(
...
-        if not updated:
-            print(
-                "Not certain if this is the latest run id. Please monitor at: "
-                + deploy_url,
-                file=self.output,
+        return self.build_no
+
+    def _get_last_deployable(self, build_no: int) -> int:
+        # This is due to optimization that some builds will not be deployable (e.g if no diff from previous build)
+        # May take a few seconds for the build to show up in the deployment list
```
jonmv, line 700: Did you actually observe that? I suppose you _could_ see a delay, but I would think that was ultra-rare.

vespa/deployment.py
```diff
@@ -550,127 +646,185 @@ def deploy(
...
             application_package=self.application_package,
             vespa_cloud_secret_token=os.environ.get(VESPA_CLOUD_SECRET_TOKEN),
         )
-        app.wait_for_application_up(max_wait=max_wait)
-        print("Finished deployment.", file=self.output)
         return app
 
+    def get_deployment_jobs(self) -> List[str]:
+        prod_regions = self.get_prod_regions()
+        return ["system-test", "staging-test"] + [
```
jonmv, line 772: This breaks down for more complex deployment specs, but those are perhaps not modelled in `pyvespa` anyway? Depending on what this is used for, it could also be gotten from the API.

vespa/deployment.py
```diff
@@ -778,7 +942,62 @@ def _set_target_cloud(self):
...
+                        # Read output from the master end of the PTY
+                        output = os.read(master, 1024).decode("utf-8")
+                        if output:
+                            print(output, end="")
+                            sys.stdout.flush()
+                        if "Success:" in output:
+                            finished = True  # Exit the loop after success message
+                            break
+
+                        # Check for input only if running in a Jupyter Notebook
```
jonmv, line 978: What happens if the CLI requests input, but none is given?

vespa/deployment.py
```diff
@@ -874,8 +1092,44 @@ def get_dev_region(self) -> str:
         return self._request("GET", "/zone/v1/environment/dev/default")["name"]
 
     def get_prod_region(self):
-        # TODO Support multiple regions
-        return self.application_package.deployment_config.regions[0]
+        regions = self.get_prod_regions()
+        return regions[0]
```
jonmv, line 1096: Any plans to support arbitrary prod regions in `get_application()`?

vespa/deployment.py
```diff
@@ -874,8 +1092,44 @@ def get_dev_region(self) -> str:
...
+            raise FileNotFoundError(
+                f"deployment.xml not found in {deployment_xml_path}. Provide disk_folder with the path to application root."
+            )
+        root = ET.fromstring(xml_data)
+
+        # Find all 'region' elements that have 'prod' as a parent
+        regions = [
+            region.text
+            for prod in root.findall(".//prod")
+            for region in prod.findall("region")
```
jonmv, line 1121: Regions aren't necessarily direct children of `prod`, but it _is_ true that a `region` will always have exactly one `prod` ancestor, and all `region` descendants of a `prod` are prod-deployment regions.

jonmv: There is no concept of `instance` here, btw.; perhaps `pyvespa` doesn't support that?

vespa/deployment.py
```diff
@@ -886,8 +1140,85 @@ def get_connection_response_with_retry(self, method, path, body, headers):
...
+                )
+            except ValueError as e:
+                print(f"Error parsing the date: {e}")
+                raise
+            if expires_at < datetime.now(timezone.utc):
+                raise Exception("Authentication failed, token is still expired.")
+
+        return auth["providers"]["auth0"]["systems"]["public"]["access_token"]
+
+    def _request_access_token(
```
jonmv, line 1186: Perhaps `_request_with_access_token`? Likewise for the one with `api_key` below.

vespa/deployment.py
```diff
@@ -973,20 +1295,31 @@ def get_endpoint(
...
-                if authMethod == auth_method:
-                    print(
-                        f"Found {auth_method} endpoint for {cluster_name}",
-                        file=self.output,
-                    )
-                    return endpoint["url"]
-        raise RuntimeError(
-            f"No {auth_method} endpoints found for container cluster {cluster_name}"
-        )
+            cluster = endpoint.get("cluster", None)
```
jonmv, line 1302: Perhaps support cluster as an optional argument to the method?

vespa/deployment.py
```diff
@@ -1044,41 +1385,59 @@ def _start_prod_deployment(self, disk_folder: str) -> None:
...
-            "X-Key": self.api_public_key_bytes,
-            "X-Authorization": standard_b64encode(signature),
-            "Content-Type": multipart_data.content_type,
-        }
-
-        response = requests.post(url, data=multipart_data, headers=headers)
-
-        message = response.json()["message"]
+        message = response.get("message", "No message provided")
+        build_no = int(response.get("build", None))
```
jonmv, line 1435: Note you could read here whether the build was identical to the previous, and if so, only then would you need to poll for the previous deployable one. It could also be nice for the user to see this in a more direct fashion. 

```json
{
  "message": "bla bla",
  "build": 123,
  "skipped": true
}
```

vespa/deployment.py
thomasht86: Good point. Reducing the level by removing "Warning" 👍

vespa/deployment.py
thomasht86: Aha, will only expose `sourceUrl`then!

vespa/deployment.py
thomasht86: Agreed! ✅

vespa/deployment.py
thomasht86: Yep, that is much more descriptive.

vespa/deployment.py
thomasht86: Added support for that now.

vespa/deployment.py
thomasht86: Deleted that function, as it was only used by the earlier approach!

vespa/deployment.py
thomasht86: Used API for this instead, and added instance-parameter.

vespa/deployment.py
jonmv: I'm just trying to understand the context. So if it's in a notebook (only), you answer "yes" for it? And otherwise, it's probably a user looking at it, that _can_ provide input?

thomasht86: No, the thing is that if it is run from a python shell the `vespa auth login` will run and behave similar to the CLI  directly, but if running from a Jupyter Notebook or Colab, it will not ask the user for input (or submit the input) without this.

thomasht86: So this is a hack to explicitly trigger a user input (when prompt is detected) and forward that input to the process.

tests/integration/test_integration_vespa_cloud_token.py
```diff
@@ -223,13 +228,17 @@ def setUp(self) -> None:
             build_status = self.vespa_cloud.check_production_build_status(
                 build_no=self.build_no
             )
-            if build_status["status"] == "done":  # TODO:  add build_status["deployed"]:
-                success = True
-                break
+            if build_status["status"] == "done":
+                if build_status["deployed"]:
+                    success = True
+                    break
```
jonmv, line 234: If `status` is `done`, then I guess there's not point in polling again, regardless of whether it's a success. So perhaps provide a dedicated message explaining that it _never_ will succeed, as opposed to the wait timing out. In both cases, a link to the deployment overview in the UI would be great, if that's readily available from `pyvespa`. (It should be, if it isn't 😝)

jonmv: Oh this was still in the test class 😆 Well, the point about not polling further still stands, although I guess this situation is unlikely to occur in the test scenario.

vespa/deployment.py
```diff
@@ -1413,7 +1399,7 @@ def _start_prod_deployment(
             headers = {
                 "X-Timestamp": timestamp,
                 "X-Content-Hash": content_hash,
-                "X-Key-Id": self.tenant + ":" + self.application + ":" + "default",
+                "X-Key-Id": self.tenant + ":" + self.application + ":" + instance,
```
jonmv, line 1402: Yeah, nice find!

vespa/deployment.py
```diff
@@ -1432,8 +1418,16 @@ def _start_prod_deployment(
...
         )
         message = response.get("message", "No message provided")
-        build_no = int(response.get("build", None))
         print(message, file=self.output)
+        build_no = int(response.get("build"))
+        skipped = bool(response.get("skipped"))
+        if skipped:
+            deployable_build_no = self._get_last_deployable(build_no)
+            print(
+                f"Build {build_no} was skipped. Returning last deployable build {deployable_build_no} instead.",
```
jonmv, line 1427:
```suggestion
                f"Build {build_no} will not be deployed, being equal to the previous one. Returning last deployable build {deployable_build_no} instead.",
```

jonmv (dismissed):
Great work!

vespa/deployment.py
thomasht86: Yeah, thats better!

tests/integration/test_integration_vespa_cloud_token.py
thomasht86: Good point! 
The link to the deployment in console is shown in `deploy_to_prod()`. 
Could probably add separate function - `wait_for_production_deployment(max_wait: int)`. As this is probably something the user will want to do anyway (and we in our tests).

jonmv: Sounds like a good idea!

jonmv (approved):

PR #819, opened 2024-07-03
Support maven-packaged deployment to prod

thomasht86:

This PR enables deployment to prod, also when an application is packaged with maven first.


kkraune (approved):
Great, let's try this!

PR #823, opened 2024-07-04
Thomasht86/bump twine and requests

### LINKED ISSUE #820: Bump requests from 2.26.0 to 2.32.2
I encountered an issue when using PyVespa v0.42.0 installed with Poetry. Attempting to import Vespa using the following code:
`from vespa.application import Vespa`

results in the following error:
ImportError: cannot import name 'JSONDecodeError' from 'requests.exceptions'

Upon investigation, I found that the JSONDecodeError class was added to requests.exceptions starting from version 2.27.0. However, your pyproject.toml file specifies requests v2.26.0, which is why this error occurs.

I suggest updating the requests dependency to v2.32.0 to resolve this issue. The same error is present in PyVespa v0.43.0 and potentially in other versions as well.

Thank you for addressing this.

thomasht86:

1. Bump version of twine (which caused release pipeline to fail due to https://stackoverflow.com/questions/78694572/unable-to-upload-pip-package-to-google-artifact-registry-using-twine-err-keyer) 
2. Pin version of `requests` for pyvespa to `>=2.32.0` to fix #820

We are using the `[dev]` version in CI. Should make sure at least unit testing is run with as little diff as possible from main dependencies (only pytest?) to avoid this from happening again.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ readme = "README.md"
 keywords = ["vespa", "search engine", "data science"]
 classifiers = ["License :: OSI Approved :: Apache Software License"]
 dependencies = [
-    "requests",
+    "requests>=2.32.0",
     "requests_toolbelt",
     "docker",
     "jinja2",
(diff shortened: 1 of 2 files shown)


kkraune (commented):
👍

PR #827, opened 2024-07-04
add source-url and make it waiting optional

thomasht86:

It should help to pass along the source-url argument. 
Also made it possible to return immediately by passing `max-wait=-1`.


kkraune (approved):
👍

PR #829, opened 2024-07-04
Simplify `getting-started-pyvespa-cloud.ipynb`

thomasht86:

Creating a PR with suggested simplification of `getting-started-pyvespa-cloud.ipynb` with support for interactive login.

Will prompt users running in notebook/Colab to login interactively. (Still works in CI/CD as api-key will be checked first if provided).  Please try running it, maybe output during/after auth is a bit too verbose? 

Removed quite a bit of the authentication explanation, maybe too much, but it looks more like a quickstart now. 
Once we agree on approach, I can batch-update rest of the cloud notebooks.


kkraune (approved):
tinier - I like it :-)

Issue #843, opened 2024-07-29
 ModuleNotFoundError: No module named 'termios'" error after pyvespa version 0.40.0 in Windows


omer38:
When I install pyvespa versions > 0.40.0, I got 
Traceback (most recent call last):
  File "C:\Users\omert\PycharmProjects\InternVenv\vespa_omer.py", line 11, in <module>
    from vespa.deployment import VespaDocker
  File "C:\Users\omert\OneDrive\Masaüstü\Example_project2\venv\Lib\site-packages\vespa\deployment.py", line 16, in <module>
    import pty
  File "C:\Users\omert\AppData\Local\Programs\Python\Python311\Lib\pty.py", line 12, in <module>
    import tty
  File "C:\Users\omert\AppData\Local\Programs\Python\Python311\Lib\tty.py", line 5, in <module>
    from termios import *
ModuleNotFoundError: No module named 'termios'. error.
It didn't give this error with earlier versions. I cannot download 'termios' on windows, so the package does not work after 0.40.0.

Issue #842, opened 2024-07-29
Add OS and python-version matrix for unit tests


thomasht86:
Currently, we only run tests on linux with Python 3.8. 
Should run at least unit tests on macOS, windows and ubuntu with Python 3.8-3.12.

PR #845, opened 2024-07-29
Raise on interactive login windows

### LINKED ISSUE #841: Fix windows-compatibility of `VespaCloud()._vespa_auth_login`
This currently uses `pty`-module, with [availability on Unix only](https://docs.python.org/3/library/tty.html).
This makes the import of the module fail for windows-users with `ImportError: No module named termios`

thomasht86:

Addresses #841, #842 and #843. 

Considered using 3rd party library like [pexpect](https://pypi.org/project/pexpect/) to try to provide support, but decided against it for now, as supporting windows in stdlib's `pty` is an [open issue in cpython](https://github.com/python/cpython/issues/85323), and is likely to be fixed in the future. 

Will now not throw `ImportError` on windows on module import, but instead raise a `NotImplementedError` only when interactive login is attempted on windows, with a workaround. 

Also added cross-platform [`macos-latest`, `ubuntu-latest`, `windows-latest`] matrix-testing for unit tests aimed to prevent cross-platform compatibility issues in the future.


.github/workflows/pyvespa-unit-tests.yml
```diff
@@ -0,0 +1,34 @@
...
+  workflow_dispatch:
+  pull_request:
+    branches:
+      - master
+
+jobs:
+  test:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
```
kkraune, line 13: hopefully the unit tests do not take much time? this is a large matrix ;-) ...

kkraune (approved):
👍

Issue #850, opened 2024-07-31
Pin `requests`-version


thomasht86:
This is causing segfaults with `requests>=2.32`- https://github.com/psf/requests/issues/6745

thomasht86:
closed by #851

PR #851, opened 2024-07-31
Thomasht86/pin requests version

thomasht86:

Ref #850. Edit version to avoid regression introduced in `requests==2.32.0`. 
This made our `feed_iterable()` fail when using cert, causing integration tests to fail with segfault.
Pinning to `2.31` fixes the issue. 
 
Waiting for fix in coming `requests`-releases.


kkraune (approved):
👍

PR #863, opened 2024-08-07
Migrated non-cloud screwdriver pipelines to GitHub Actions workflows

glebashnik:

A couple of things worth mentioing:
1. It uses https://github.com/jlumbroso/free-disk-space action to free disk space in ubuntu-latest runner.
2. Notebooks are now executed using papermill for better logging.

A lot of commits since we can't test the workflows locally.
I would like to squash them before merge.
Also need to update triggers after review and remove corresponding screwdriver pipelines.


.github/workflows/docs-test.yml
```diff
@@ -0,0 +1,41 @@
...
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'
+          cache: 'pip'
+          cache-dependency-path: |
+            pyproject.toml
+
+      - name: Install dependencies
+        run: |
+          pip install -e .[dev]
```
thomasht86, line 30:
```suggestion
          pip install -e .[dev,docs]
```

thomasht86 (dismissed):
Great! Think this is almost ready to be merged now. Just commented out screwdriver.yml, to avoid that one resulting in a failed job.

thomasht86 (approved):
Nice! 
44 ✅ tests 🙌

PR #865, opened 2024-08-08
Glebashnik/move all requirements to pyproject

glebashnik:


thomasht86:
That reminds me - we also need to create a separate workflow for linkcheck and building docs - replacing these:
https://github.com/vespa-engine/pyvespa/blob/afb8f61970fa014495bc34664d379982d6a5667c/screwdriver.yaml#L35-L38

PR #870, opened 2024-08-12
fix: Allows symlink when loading app pkg from disk

glesperance:
I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner. [x]

---

When invoking VespaDocker().deploy_from_disk pyvespa is not following symlinks.
This fixes that by turning passing followlinks to os.walk(".", followlinks=True).


thomasht86:
Thanks @glesperance ! 
We are currently migrating our CI pipelines, and have yet to configure the process of running workflows on PRs from forks (that need access to secrets). 
Will return to this as soon as we have this sorted out.
 
Thanks for your patience!

thomasht86 (approved):
Thank you for your contribution and patience.

PR #878, opened 2024-08-16
chore(CI): change workflow trigger rules

esolitos:
## What

- Adding "push: {default-branch}" for testing final integration of changes with mainline
- Remove schedule from GH Actions workflows where is not providing any additional value
- Updated schedule to be weekly (Sunday) where remains useful

## Why

This streamlines the tests to run when is necessary and avoid unnecessary additional runs.

---


thomasht86 (approved):
Great! 👍

PR #881, opened 2024-08-20
md to html table

thomasht86:

Seems like the markdown table did not render properly on readthedocs. Probably needs [this extension](https://pypi.org/project/sphinx-markdown-tables/). 
Converted to html instead, to avoid dep.


kkraune (approved):
👍

Issue #872, opened 2024-08-13
Add compression support


jobergum:
Adding accept-encoding to the request will make Vespa compress the response, which can reduce latency for cross-region query traffic: 

```Accept-Encoding: gzip```

thomasht86:
This is the default behavior for both `httpx` and `requests`. 
See
They will also be decompressed automatically.
See #883 
- From [httpx docs](https://www.python-httpx.org/quickstart/):
> Any gzip and deflate HTTP response encodings will automatically be decoded for you.
- From [requests FAQ](https://requests.readthedocs.io/en/latest/community/faq/):
> Requests automatically decompresses gzip-encoded responses, and does its best to decode response content to unicode when possible.

PR #883, opened 2024-08-21
Add integration test to test for accept-encoding: gzip

thomasht86:

Ref #872.

It turns out both `requests` and `httpx` adds a default header with `Accept-Encoding: gzip, deflate`. 
The Vespa container also replies with `content-encoding: gzip`

The behavior can be verified with

```python
import requests
import httpx

# Using requests
requests_response = requests.get('https://httpbin.org')
print("Requests - Accept-Encoding header:", requests_response.request.headers.get('Accept-Encoding'))

# Using httpx
httpx_response = httpx.get('https://httpbin.org')
print("HTTPX - Accept-Encoding header:", httpx_response.request.headers.get('Accept-Encoding'))
```

They will also be decompressed automatically.
See:

- From [httpx docs](https://www.python-httpx.org/quickstart/):
> Any gzip and deflate HTTP response encodings will automatically be decoded for you.
- From [requests FAQ](https://requests.readthedocs.io/en/latest/community/faq/):
> Requests automatically decompresses gzip-encoded responses, and does its best to decode response content to unicode when possible.

This PR adds integration tests to test for this.

### DIFF
--- a/tests/integration/test_integration_docker.py
+++ b/tests/integration/test_integration_docker.py
@@ -1135,7 +1135,7 @@ class TestQaApplication(TestApplicationCommon):
         self.app.delete_all_docs(content_cluster_name="qa_content", schema="sentence")
 
     def test_sync_client_accept_encoding(self):
-        self.sync_client_accept_encoding(app=self.app)
+        self.sync_client_accept_encoding_gzip(app=self.app)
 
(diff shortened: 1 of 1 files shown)


jobergum (approved):

jobergum:
Thanks for investigating this

PR #886, opened 2024-08-22
Switch to new control-plane API

thomasht86:

Took the opportunity to clean up a bit for a more consolidatet http-client usage. 
Think this will also remove instability in integration tests (where http.connection was used previously)

- [x] Use new control plane-API
- [x] Hardcode dev region (similar to vespacli)
- [x] Switch to httpx-client for more unified usage
- [x] Add common user-agent header for all clients with version
- [x] Add integration tests for user-agent header

### DIFF
--- a/.github/workflows/notebooks-cloud.yml
+++ b/.github/workflows/notebooks-cloud.yml
@@ -9,7 +9,7 @@ on:
     branches:
       - master
   schedule:
-   - cron: '0 11 * * 0'
+    - cron: "0 11 * * 0"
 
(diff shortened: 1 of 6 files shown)


thomasht86:
This should be good now. 
Spent some time making the tests pass, and think I was able to fix what caused sporadic integration test failures. 
Now, all the http-connections used everywhere are subjected to the retry-policies, and use either `requests` or `httpx`

tokle (approved):

PR #894, opened 2024-08-28
infer auth method without request

thomasht86:

While investigating query performance, I noticed that we perform HTTP-requests to infer auth_method. This results in multiple, unnecessary HTTP-requests before the "actual request" when initializing a `VespaSync`/`VespaAsync`-connection, leading to higher response times.

This PR eliminates the extra calls, and instead infers `auth_method` from the passed parameters.


thomasht86:
Also triggered cloud tests manually ✅
https://github.com/vespa-engine/pyvespa/actions/runs/10594808080

kkraune (approved):
👍

PR #897, opened 2024-08-29
Revert "infer auth method without request"

esolitos:
## What

Reverts vespa-engine/pyvespa#894

## Why

We noticed some issues on authenticating with mtls to recently deployed applications on this version. 
Still not sure if this is a real issue, as it was observed locally, but not in integration tests, or Colab.


thomasht86 (approved):
🙏

Issue #905, opened 2024-09-04
Ensure notebooks-cloud integration tests run on **current** pyvespa-version


thomasht86:
Currently, since we `pip3 install pyvespa` inside the notebooks after `pip install -e .[dev,notebooks]` in the action workflow, the former will override pyvespa with latest version from pypi, rather than the checked out version.

thomasht86:
Closed by #907

Issue #824, opened 2024-07-04
Lock main dependencies in CI


thomasht86:
Ref #820. 
We should update our CI-pipeline, which are run with `[dev]` so that we have locked version of main dependencies, to make sure our tests are run with the exact same versions as the dependencies. This will make sure tests fail on similar issues in the future.

thomasht86:
closed by #907

PR #907, opened 2024-09-05
(ci) fix notebook environments in ci

thomasht86:

This PR solves #905 while still allowing us to keep using

```bash
!pip3 install package1 package2
```
in the notebooks, so they they can be run directly from Colab. 

To achieve this, I added a script that:

- Identifies these lines.
- Removes `pyvespa` and `vespacli` as they are already installed. 
- Adds the other packages to a `additional_requirements.txt` 
- Install these only. 
- Comments out the installation line from the notebook
- Runs the notebook test. 

Also, moved to using [uv](https://docs.astral.sh/uv/) for package management for speed and ability to lock versions, in preparation for #824 . (For this workflow only for now, others to follow)

### DIFF
--- a/.github/scripts/replace_pip_install_notebooks.sh
+++ b/.github/scripts/replace_pip_install_notebooks.sh
@@ -1,0 +1,62 @@
+#!/bin/bash
+
+# Script to extract and modify pip install commands from a Jupyter notebook
+# Example: If a notebook contains the following line:
+# !pip3 install -U pyvespa sentence-transformers vespacli numpy
+# The script will extract the package names, excluding pyvespa and vespacli (already installed) and save the new packages to additional_requirements.txt
+# The resulting additional_requirements.txt will contain:
+# sentence-transformers 
+# numpy
+# It will also comment out the original line in the notebook
+
+# Check if an argument was provided
+if [ "$#" -ne 1 ]; then
+    echo "Usage: $0 path_to_notebook"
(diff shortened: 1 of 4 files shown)


.github/scripts/replace_pip_install_notebooks.sh
```diff
@@ -0,0 +1,53 @@
...
+if [ "$#" -ne 1 ]; then
+    echo "Usage: $0 path_to_notebook"
+    exit 1
+fi
+
+# Assign the first argument to the variable notebook
+notebook=$1
+
+# Now you can use the variable $notebook in your script
+echo "The path to the notebook is \"$notebook\""
```
esolitos, line 13: Perhaps some (most?) of those echo should use [the debug output option/prefix](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-a-debug-message), so that we only show what is necessary on a normal run?

esolitos (approved):
Tests are green and the logic seems fine. 😊 

Added just a couple of comments, they don't need to be done now, just a suggestion fro the future.

.github/scripts/replace_pip_install_notebooks.sh
thomasht86: Yes, didn't know about this. Thx! 
Will fix the next times these files are touched!

Issue #887, opened 2024-08-22
Add compression support for feed and query (post)


jobergum:
With large floats in query tensors or in documents, compressing the data can help on network time.

PR #911, opened 2024-09-12
(feat)Add compression for POST/PUT requests (sync)

thomasht86:

Ref #887. 
This PR adds compression support for synchronous POST and PUT requests by addiing `compress`-arg to `VespaSync`. 
Will enhance performance of larger requests (especially with repeating text, in eg. yql). The gains are less with tensors, due to less compression ratio. Will be increasingly beneficial if network latency to Vespa server is slow. 

The performance gain of `feed_iterable` in https://pyvespa.readthedocs.io/en/latest/examples/feed_performance_cloud.html is approximate 2x in [test run on this PR](https://github.com/vespa-engine/pyvespa/actions/runs/10828798924/job/30044993243). (~60 docs/s -> ~120 docs/s. )

The default is `"auto"`, which compresses body content larger than 1024 bytes. (Vespa CLI defaults to 512. Some simple test cases showed little gain when also considering some overhead for doing the compression). 

```python
class VespaSync(object):
    def __init__(
        self,
        app: Vespa,
        pool_maxsize: int = 10,
        pool_connections: int = 10,
        compress: Union[str, bool] = "auto",
    ) -> None:
        """
        Class to handle synchronous requests to Vespa.
        This class is intended to be used as a context manager.

        Example usage::

                with VespaSync(app) as sync_app:
                    response = sync_app.query(body=body)
                print(response)

        Can also be accessed directly through :func:`Vespa.syncio` ::

                app = Vespa(url="localhost", port=8080)
                with app.syncio() as sync_app:
                    response = sync_app.query(body=body)

        See also :func:`Vespa.feed_iterable` for a convenient way to feed data synchronously.

        Args:
            app (Vespa): Vespa app object.
            pool_maxsize (int, optional): The maximum number of connections to save in the pool. Defaults to 10.
            pool_connections (int, optional): The number of urllib3 connection pools to cache. Defaults to 10.
            compress (Union[str, bool], optional): Whether to compress the request body. Defaults to "auto", which will compress if the body is larger than 1024 bytes.
        """
```
- [x] Added unit tests
- [x] Added "integration" tests - tests that the args work. Hard to test actual compression without patching and mocking.

### DIFF
--- a/docs/sphinx/source/query.ipynb
+++ b/docs/sphinx/source/query.ipynb
@@ -1,610 +1,791 @@
 {
-    "cells": [
-        {
-            "cell_type": "markdown",
-            "metadata": {},
-            "source": [
-                "<picture>\n",
-                "  <source media=\"(prefers-color-scheme: dark)\" srcset=\"https://assets.vespa.ai/logos/Vespa-logo-green-RGB.svg\">\n",
-                "  <source media=\"(prefers-color-scheme: light)\" srcset=\"https://assets.vespa.ai/logos/Vespa-logo-dark-RGB.svg\">\n",
-                "  <img alt=\"#Vespa\" width=\"200\" src=\"https://assets.vespa.ai/logos/Vespa-logo-dark-RGB.svg\" style=\"margin-bottom: 25px;\">\n",
-                "</picture>\n",
-                "\n",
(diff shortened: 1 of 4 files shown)


jobergum (approved):
Great!

PR #917, opened 2024-09-16
do not fail if response is not json (e.g. xml)

mkevac:
This PR fixes a bug where non-JSON (e.g. XML) responses will cause failure.
Sister function `_request_with_access_token` has no such bug.


thomasht86:
Applicable tests pass. 
Thanks a lot @mkevac. Aiming to release a new version with this soon.

thomasht86 (approved):

PR #919, opened 2024-09-17
(docs) Make sure notebook is committed with output 

thomasht86:

If not, it is run (and will fail) on readthedocs, and give error output:
image

Source: https://nbsphinx.readthedocs.io/en/0.2.15/pre-executed.html


esolitos (approved):
👍🏼  Lots of changes, but seems to make sense. 😊

PR #928, opened 2024-09-19
Demonstrate that hamming MaxSim is a very good alternative to float based MaxSim 

jobergum:


thomasht86:
Seems like a very attractive tradeoff! 💯 
Would be nice if you could use a recent notebook as basis (without all the CLI-commands and auth-stuff that we now no longer need.) And Colab-badge. :)

thomasht86:
See that CI fail due to missing PIL - `pip install pillow`

jobergum:
Will do! One issue here with having this in CD is that it takes 20 minutes to encode the corpus, but let us see first how long it takes, and then maybe run it once in a while or something.

thomasht86:
Looks like we're not able to run this in CI without larger runners. 
Crashes in Colab as well with the standard CPU runtime (works with T4 though).
I'll follow up on getting us access to larger runners, but we'll probably need to exclude it for now.

jobergum:
Yeah, as the model is 12GB and it encodes the images + overhead I do imagine we pass the runner memory limits. I would like to keep this notebook though so if you could disable it from test for now that would be great. Thanks!

thomasht86 (approved):
👍

PR #932, opened 2024-09-24
(fix) Thomasht86/move relaxng files

thomasht86:

The relaxng-files are used in pyvespa module for validation, and should thus not be located in `testfiles/`-directory. 
This also caused issues with docs-building, as the testfiles are not available. 
Thanks for pointing this out.


glebashnik (approved):
Looks good

PR #915, opened 2024-09-16
(feat)First step towards full xml-configuration support

thomasht86:
I confirm that this contribution is made under the terms of the license found in the root directory of this repository's source tree and that I have the authority necessary to make this contribution on behalf of its copyright owner.

**Why**

Adding support for tags far down in the tree, ref #895, results in a lot of code, and with our current approach of using jinja-templates requires a lot of effort each time. 

This PR is in preparation to allow for supporting any `services.xml` configuration from pyvespa, with [the approach used by FastHTML](https://github.com/fastai/fastcore/blob/master/fastcore/xml.py) as inspiration. 

**How** 

By adopting the approach in this PR, we only need to provide a list of valid xml-tags for each configuration file, and all the classes will be auto-generated. 

E.g. to express the following `services.xml`:

```xml
<?xml version="1.0" encoding="utf-8" ?>
<services version='1.0' xmlns:deploy="vespa" xmlns:preprocess="properties">

  <container id='default' version='1.0'>
    <nodes count='1'/>
    <component id='ai.vespa.examples.Centroids' bundle='billion-scale-image-search'/>
    <component id='ai.vespa.examples.DimensionReducer' bundle='billion-scale-image-search'/>
    <component id="ai.vespa.examples.BPETokenizer" bundle='billion-scale-image-search'>
      <config name="ai.vespa.examples.bpe-tokenizer">
        <contextlength>77</contextlength>
        <vocabulary>files/bpe_simple_vocab_16e6.txt.gz</vocabulary>
      </config>
    </component>
    <model-evaluation>
      <onnx>
        <models>
          <model name="text_transformer">
            <intraop-threads>1</intraop-threads>
          </model>
          <model name="vespa_innerproduct_ranker">
            <intraop-threads>1</intraop-threads>
          </model>
        </models>
      </onnx>
    </model-evaluation>
    <search>
      <chain id='default' inherits='vespa'>
        <searcher id='ai.vespa.examples.searcher.DeDupingSearcher' bundle='billion-scale-image-search'/>
        <searcher id='ai.vespa.examples.searcher.RankingSearcher' bundle='billion-scale-image-search'/>
        <searcher id="ai.vespa.examples.searcher.CLIPEmbeddingSearcher" bundle="billion-scale-image-search"/>
        <searcher id='ai.vespa.examples.searcher.SPANNSearcher' bundle='billion-scale-image-search'/>
      </chain>
    </search>
    <document-api/>
    <document-processing>
[...]

Issue #854, opened 2024-07-31
Support document expiry using pyvespa


marcosestevewhalar:
Enable https://docs.vespa.ai/en/documents.html#document-expiry in pyvespa schemas

https://vespatalk.slack.com/archives/C038YTM5SUQ/p1722430339533689

PR #936, opened 2024-09-24
(feat) Support document expiry

thomasht86:

**Why** 

Ref. #854. 

**What** 

- [x] Added unit tests and integration tests, as well as notebook demonstrating how this can be done using the new `ServicesConfiguration`-approach. 
- [x] Added more explanations ref feedback on #915.
- [x] More robust `validate_services`-function (with tests). 

The intention is to add more examples to the `advanced-configuration.ipynb` as we go.


tests/unit/test_package.py
```diff
@@ -1805,3 +1806,83 @@ def test_default_service_config_to_text(self):
...
+                        indexing=["attribute", "summary"],
+                        attribute=["fast-access"],
+                    ),
+                ]
+            ),
+        )
+        # Create a ServicesConfiguration with document-expiry set to 1 day (timestamp > now() - 86400)
+        services_config = ServicesConfiguration(
+            application_name=application_name,
+            services_config=services(
```
glebashnik, line 1840: Starting here linters complain that services, container, etc are not defined.
They can't see dynamically generated imports.

glebashnik (commented):
Looks good, see the comment about problem with linters.

glebashnik (approved):

tests/unit/test_package.py
thomasht86: Ref offline chat, I consider this as resolved for now, and taking a look on whether we can support this in the future. 
When using linter-settings as defined in`pyproject.toml` these errors are ignored.

PR #944, opened 2024-10-01
ColQwen2 example

jobergum:
Demonstrates using the new model checkpoint based on QWen2 https://huggingface.co/vidore/colqwen2-v0.1


thomasht86:
Awesome with notebook with better and faster model! 🚀 

Linkcheck will pass as the docs are updated. 
PR is open to fix Notebooks except cloud. 

A few points to consider:
- Rename application to `visionrag6` to avoid conflict in naming (will fail if run in CI in parallel)
- Add `tqdm` to pip installs (may be depended on by one of the others, so possibly not necessary)
- Add to TOC

thomasht86 (approved):
Great!

Issue #950, opened 2024-10-09
pyvespa 0.46.0+ total_timeout is not passed


dzianis-kryvapusk:
The issue exists since 0.46.0.
I have the following code to invoke the query
```
v_client = Vespa(settings.endpoint, cert=settings.vespa_cert_file)
async with v_client.asyncio(total_timeout=300) as async_client:
     req['timeout'] = '300s',
     resp = await async_client.query(body=req)
```
When the query runs for longer than 10sec it fails with the following error:
`RetryError[<Future at 0x1b1c93b70a0 state=finished raised ReadTimeout>`
To reproduce the issue, try to use the following searcher:
```
@After(MinimalQueryInserter.EXTERNAL_YQL)
public class SleepSearcher extends Searcher {
    @Override
    public Result search(Query query, Execution execution) {
        Result result = execution.search(query);

        Integer timeToSleep = getValue(query);
        if (timeToSleep != null) {
            try {
                TimeUnit.SECONDS.sleep(timeToSleep);
            } catch (InterruptedException e) {
                return result;
            }
        }
        return result;
    }

    private Integer getValue(Query query) {
        try {
            return Integer.parseInt(query.properties().getString(new CompoundName("sleep")));
        } catch (NullPointerException | NumberFormatException e) {
            return null;
        }
    }
}
```
Then, pass `req['sleep'] = '20'`.

thomasht86:
This is now fixed in ltest release (`0.50.0`).
Thanks for reporting.

PR #953, opened 2024-10-16
(bugfix) Fix asyncclient timeout param

thomasht86:

Closing issue #950. 
✅ Added unit tests to test that params are correctly passed through.
Also added `kwargs` that will be passed to async-client.


vespa/application.py
```diff
@@ -1517,9 +1543,17 @@ def _open_httpx_client(self):
         limits = httpx.Limits(
             max_keepalive_connections=self.connections,
             max_connections=self.connections,
-            keepalive_expiry=10,  # This should NOT exceed the keepalive_timeout on the Server, otherwise we will get ConnectionTerminated errors.
+            keepalive_expiry=self.keepalive_expiry,  # This should NOT exceed the keepalive_timeout on the Server, otherwise we will get ConnectionTerminated errors.
         )
-        timeout = httpx.Timeout(pool=5, connect=5, read=5, write=5)
+        if self.total_timeout:
+            timeout = httpx.Timeout(self.total_timeout)
```
glebashnik, line 1549: If I understand the following code correctly this is not really total_timeout but more like default value for different timeouts:
https://github.com/encode/httpx/blob/1bf1fc0ea8417c8e539914aeafdb48f15591ba75/httpx/_config.py#L246
Perhaps we can let users pass Timeout object so we don't have to deal with the logic of how to set timeouts? This does break abstraction layer so kind of a tradeoff.

tests/unit/test_application.py
```diff
@@ -693,5 +696,203 @@ def test_retry_on_429_status(self):
...
+    async def test_read_timeout(self):
+        app = Vespa(url="http://localhost", port=8080)
+        read_timeout = 1  # seconds
+        vespa_async = VespaAsync(app=app, total_timeout=None, read_timeout=read_timeout)
+
+        with patch(
+            "vespa.application.httpx.AsyncClient", autospec=True
+        ) as MockAsyncClient:
+            mock_client_instance = MockAsyncClient.return_value
+            mock_client_instance.post = AsyncMock(
```
glebashnik, line 739: Do we have a better way to simulate timeout than mock throwing an exception directly?

glebashnik (changes requested):

vespa/application.py
thomasht86: Yes, good point. Tried to clarify in docstring for `total_timeout`-parameter. 
Did it this way to ensure backwards compatibility (by not changing parameter name). 
Good suggestion to let user pass Timeout. 
My suggestion is then to deprecate `total_timeout` parameter (issue `DeprecationWarning`,  and introduce `timeout: httpx.Timeout = httpx.Timeout(5)`. 
Does that sound reasonable?

glebashnik: Sounds good.

tests/unit/test_application.py
thomasht86: Hmm, agree that would be nice. I'll try to see if I can find a more relevant way.

glebashnik (approved):
Looks good

PR #966, opened 2024-11-05
remove non-text fields

andreer:
Made this change directly in the ipynb from github, let's see if it still works!


thomasht86 (approved):
Great!

PR #971, opened 2024-11-12
(ci) fix vespacli workflow

thomasht86:

Wait with merge until we can test with latest CLI release.

### DIFF
--- a/.github/workflows/release-vespacli.yml
+++ b/.github/workflows/release-vespacli.yml
@@ -37,7 +37,6 @@ jobs:
           # Set the output variable
           echo "version=$version" >> $GITHUB_OUTPUT
 
-
   check-and-update:
     runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


vespacli/pyproject.toml
```diff
@@ -8,7 +8,7 @@ description = "A Python wrapper for Vespa CLI tools, supporting multiple platfor
 keywords = [ "vespa", "cli", "wrapper",]
 name = "vespacli"
 readme = "README.md"
-version = "8.dev"
+version = "8.0.dev"
```
esolitos, line 11: This is not correct, though? 🤔
If we are mapping the version with VespaCLI, 8.0.x is not correct, while 8.x it is.

thomasht86: Changed back

esolitos (approved):

PR #975, opened 2024-11-19
adjust var name to not override builtin

regmibijay:

Currently the code block in provided notebook 

```python3
if torch.cuda.is_available():
    device = torch.device("cuda")
    type = torch.bfloat16
elif torch.backends.mps.is_available():
    device = torch.device("mps")
    type = torch.float32
else:
    device = torch.device("cpu")
    type = torch.float32
```
overrides builtin function `type`. This PR renames the portion to use `dtype` instead of type to address this issue.


thomasht86:
As you can see in #979, we renamed this variable in several notebooks. Thanks a lot for reporting. :)

regmibijay:
thanks for the heads up Thomas. Awesome stuff.

PR #981, opened 2024-11-22
(ci) fix integrationtest vectorsearch

thomasht86:

This integration test has been unstable and slooow. 🐌 
Some updates to improve stability and speed. 

- [x] reuse common environment and run workflows in parallel
- [x] comment out prod deployment (not only skip)
- [x] switch to Python 3.10

PR run here: https://github.com/vespa-engine/pyvespa/actions/runs/11970588855

### DIFF
--- a/.github/workflows/integration-cloud.yml
+++ b/.github/workflows/integration-cloud.yml
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - thomasht86/fix-integrationtest-vectorsearch
   schedule:
     - cron: "0 11 * * 0"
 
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-cloud.yml
```diff
@@ -13,20 +13,58 @@ concurrency:
...
+
+      - name: Set up Python
+        id: setup-python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+
+      - name: Cache dependencies
+        id: cache-python
+        uses: actions/cache@v3
```
esolitos, line 31:
```suggestion
        uses: actions/cache@v4
```

.github/workflows/integration-cloud.yml
```diff
@@ -13,20 +13,58 @@ concurrency:
...
+      - name: Install dependencies
+        if: steps.cache-python.outputs.cache-hit != 'true'
+        run: |
+          pip install -e .[dev]
+
+      - name: Upload Python environment
+        uses: actions/upload-artifact@v3
+        with:
+          name: python-environment
+          path: ~/.cache/pip
```
esolitos, line 45: I don't understand why you would use this instead of simply rely on `actions/cache` for restoring as well?

.github/workflows/integration-cloud.yml
```diff
@@ -13,20 +13,58 @@ concurrency:
...
-          cache: "pip"
-          cache-dependency-path: |
-            pyproject.toml
+          python-version: "3.10"
+
+      - name: Download Python environment
+        uses: actions/download-artifact@v3
+        with:
+          name: python-environment
+          path: ~/.cache/pip
```
esolitos, line 62: Use actions/cache + `restore-key` instead?

esolitos (commented):
I'm not sure why you wouldn't reuse the cache and instead upload it as artifact?

What I would typically in this situation is one of those:

a. Have a "build" job which builds the application and stores an actions/upload-artifact. The dependant jobs only get the built artifact and don't run any install steps.

b. Don't have a build job and simply use actions/cache with `restore-keys` that can fallback on a "older version" of the caches perfect match fails.

You seem to have mixed up caches and artifact and I don't see the benefit in this approach, unless I am missing something, that is. 😅

.github/workflows/integration-cloud.yml
esolitos: Also, outdated version:
```suggestion
      - name: Download Python environment
        uses: actions/download-artifact@v4
        with:
          name: python-environment
          path: ~/.cache/pip
```

thomasht86:
Thanks! I will have another look at the `cache`-action to see if we can skip the artifact upload/download!

thomasht86:
Ok, so it is 2 different things we ideally would want 2 cache. 
1. Pip packages directory. This  is now handled automatically by `setup-python`-action. 
2. It would also be nice to cache the generated environment after installing pyvespa from source. This would probably require a more advanced caching, but not so much time to save on this, so leaving that for now.

.github/workflows/integration-cloud.yml
```diff
@@ -5,6 +5,7 @@ on:
   push:
     branches:
       - master
+      - thomasht86/fix-integrationtest-vectorsearch
```
esolitos, line 8: You can prevent the need to set a branch for testing this by using the `pull_request` and filtering by `paths` 😊 
 
I believe this would be a good addition, I started using this trick myself. 😊 

```suggestion
  pull_request:
    paths: [".github/workflows/integration-cloud.yml"]
```

_(This will trigger for PRs **only** if the workflow is modified)_

esolitos (approved):
👍🏼

PR #984, opened 2024-12-03
(ci) Thomasht86/fix slow colpali notebooks

thomasht86:

Seems like using `torch.autocast()` does not work well when running on cpu - https://github.com/pytorch/pytorch/issues/118499.

These notebooks were 2 oom slower to do inference of the colpali model.  

- [x] Removed autocast and also upgraded to use `colpali_engine==3.1.0`.
- [x] Also added a cleanup cell to delete vespa application if running in CI. 
- [x] Tested that they run on Colab with T4 GPU. 

Fixed a broken link in M3-notebook (large diff due to ruff-formatting)


jobergum (approved):
Wonderful!

PR #985, opened 2024-12-04
(feat) Query builder

thomasht86:

Consider this a first draft, and I can go through it tomorrow. 

To discuss: 
- Should we allow for using `kwargs` for annotations - good for some, but many annotations are camelCase - does not good in python. 
- Rename `Q` and `G`- to what? 

TODO:

- [x] Integration test for grouping


thomasht86:
TODOs after review: 

- [ ] Rename `querybuilder/main.py`. Check if there is a convention. If not, `querybuilder.py`
- [ ] `Queryfield` -> `QueryField`
- [ ] maybe use `qb` condition.
- [ ] Describe how to build many queries programmatically (need to copy) 
- [ ] snake case for eg groupBy - dotProduct. 
- [ ] grouping to its own module

glebashnik (approved):
Looks very good, both design wise and tests.
I setup env and ran tests in Python 2.12 (ops, meant 3.13).
For future improvement, it would be great to be able to run tests in different Python versions. Tools like Nox can help with this.

thomasht86:
> Looks very good, both design wise and tests. I setup env and ran tests in Python 2.12. For future improvement, it would be great to be able to run tests in different Python versions. Tools like Nox can help with this.

Thanks. Assume you mean 3.12? 😅 We do run unit tests on matrix of Python 3.8-3.12 and macos, ubuntu, windows. 
Doing it for all would take a lot of time and cost a lot, but maybe we could do some of the integration tests at least.

PR #987, opened 2024-12-10
Fix type annotations in `feed_data_point`

kevinhu:
In `feed_data_point`, `namespace`, `groupname`, and `semaphore` should be annotated as Optional, as they are `None` by default.

### DIFF
--- a/vespa/application.py
+++ b/vespa/application.py
@@ -1651,9 +1651,9 @@ class VespaAsync(object):
         schema: str,
         data_id: str,
         fields: Dict,
-        namespace: str = None,
-        groupname: str = None,
-        semaphore: asyncio.Semaphore = None,
+        namespace: str | None = None,
+        groupname: str | None = None,
+        semaphore: asyncio.Semaphore | None = None,
         **kwargs,
     ) -> VespaResponse:
         path = self.app.get_document_v1_path(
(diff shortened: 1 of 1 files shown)


thomasht86:
Thanks - good catch! We need to use `namespace: Optional[str] = None` for now though, as we need to support Python 3.9 for a while longer.

vespa/application.py
```diff
@@ -1651,9 +1651,9 @@ async def feed_data_point(
         schema: str,
         data_id: str,
         fields: Dict,
-        namespace: str = None,
-        groupname: str = None,
-        semaphore: asyncio.Semaphore = None,
+        namespace: str | None = None,
+        groupname: str | None = None,
+        semaphore: asyncio.Semaphore | None = None,
```
thomasht86, line 1656:
```suggestion
        namespace: Optional[str] = None,
        groupname: Optional[str] = None,
        semaphore: Optional[asyncio.Semaphore] = None,
```

thomasht86 (changes requested):

vespa/application.py
thomasht86: This syntax was [introduced in Python 3.10](https://peps.python.org/pep-0604/), so we can't switch to it yet unfortunately.

PR #1000, opened 2025-01-13
Thomasht86/querybuilder documentation

thomasht86:

Updated the `query`-notebook with a 5 examples of using the querybuilder API to construct queries against live docsearch app.


kkraune (approved):
nice! good to have varied query use cases, including grouping

PR #1002, opened 2025-01-21
Thomasht86/add unittest matches

thomasht86:

Adding a test for matches with regex, to reply to https://stackoverflow.com/questions/79372000/escaping-special-characters-in-a-vespa-yql-matches-query


glebashnik (approved):
Looks good

PR #1010, opened 2025-02-10
Fix notebooks except cloud test

thomasht86:

This fails if docs are updated more recent than 1 hour, can set to now() instead.


esolitos (commented):
> "We want to find the documents where one of the indexed fields contains the query term `embedding`,is updated after Jan 1st 2024 and 1 hour ago, and have the documents ranked the 'documentation' rank profile. See https://github.com/vespa-cloud/vespa-documentation-search/blob/main/src/main/application/schemas/doc.sd. "

We should either update the test above, or understand why this started failing if it used to be ok?

esolitos (approved):

PR #1017, opened 2025-02-18
Support for rank-score-drop-limit in global phase

dainiusjocas:

The [support](https://github.com/vespa-engine/vespa/pull/33298) for the `rank-score-drop-limit` in the `global-phase` is in the works. Let's support it in pyvespa.


thomasht86 (approved):
Thanks!

PR #1021, opened 2025-02-21
make gpu example work

andreer:
If I don't do this it generates two completely separate tags which aren't accepted

### DIFF
--- a/vespa/package.py
+++ b/vespa/package.py
@@ -134,7 +134,7 @@ class Summary(object):
                 if isinstance(field[1], str):
                     tmp_string += f"{field[1]}"
                 else:
-                    tmp_string += f'{", ".join(field[1])}'
+                    tmp_string += f"{', '.join(field[1])}"
                 final_list.append(tmp_string)
 
         final_list.append("}")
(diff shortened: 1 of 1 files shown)


thomasht86:
Hmm..Which example are you trying to run? Doctest failing now, but it seems wrong. 😅 
```
        Example:

        >>> ContainerCluster(id="example_container",
        ...    nodes=Nodes(
        ...        count="2",
        ...        parameters=[
        ...            Parameter("resources", {"vcpu": "4.0", "memory": "16Gb", "disk": "125Gb"}),
        ...            Parameter("gpu", {"count": "1", "memory": "16Gb"}),
        ...            Parameter("node", {"hostalias": "node1", "distribution-key": "0"}),
        ...        ]
        ...    )
        ... )
        ContainerCluster(id="example_container", version="1.0", nodes="Nodes(count="2")")
 ```
       
I would recommend to rather use the advanced configuration for setting this (lowercased tags) - example: https://github.com/vespa-engine/pyvespa/blob/e085a7222d79f0bf8347f190c25178535c77579e/tests/unit/test_configuration.py#L544
  
But we should update the docstring at least.

andreer:
Yeah I was literally just trying to copy the code from the docstring into my own WIP pyvespa app to get a gpu but it doesn't work - it creates two separate xml tags like
`<resources vcpu=""... />`
`<gpu count="1" ... />`
Which is not valid (the gpu tag needs to be inside the resources tag). I have no idea why fixing the doc string would make something fail 😄 but it would be good to have a working example in the docs instead of this broken one.

thomasht86 (approved):
With an added comma it worked 😄

PR #1023, opened 2025-02-24
fix: rendering of rank-score-drop-limit

dainiusjocas:

It turns out that I forgot to adjust the schema template so that `rank-score-drop-limit` would be properly rendered. 

Also, in Jinja templates if value is 0 then it is considered falsy, and therefore `{% if value.second_phase.rank_score_drop_limit %}` renders nothing. 
I've adjusted the check to `{% if value.second_phase.rank_score_drop_limit != None %}` so that 0 values are still rendered.


thomasht86:
Updated tests with simpler syntax. Passes now.
Thanks a lot! 💪 🙌

thomasht86 (approved):

PR #1036, opened 2025-03-05
Support `slice_id`in visit request

thomasht86:

### DIFF
--- a/vespa/application.py
+++ b/vespa/application.py
@@ -857,6 +857,7 @@ class Vespa(object):
         slices: int = 1,
         selection: str = "true",
         wanted_document_count: int = 500,
+        slice_id: int = -1,
         **kwargs,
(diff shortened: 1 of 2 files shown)


vespa/application.py
```diff
@@ -857,6 +857,7 @@ def visit(
         slices: int = 1,
         selection: str = "true",
         wanted_document_count: int = 500,
+        slice_id: int = -1,
```
glebashnik, line 860: Consider Optional with None default value instead of -1 for more pythonic experience.

glebashnik (approved):
Minor suggestion.

glebashnik (dismissed):

vespa/application.py
thomasht86: good point! will update

glebashnik (approved):
Looks good.

PR #1043, opened 2025-03-25
Migration to MkDocs: docstrings translation

vadimkatsel:
Important:

- All of the configurations of mkdocs are presented in mkdocs.yml

- To run it locally you need:

1.  pip install the following(all of them already pre-installed on venv, so that `source venv/bin/activate` as an option):
[mkdocs-material](https://squidfunk.github.io/mkdocs-material/getting-started/): `pip install mkdocs-material`
[mkdocs-jupyter](https://github.com/danielfrg/mkdocs-jupyter): `pip install mkdocs-jupyter`
[mkdocstrings](https://mkdocstrings.github.io/): `pip install 'mkdocstrings[python]'`
[mkdocs-autoapi](https://mkdocs-autoapi.readthedocs.io/en/latest/): `pip install mkdocs-autoapi`

2.  run mkdocs:
`mkdocs serve `
for debug:
`mkdocs serve --verbose `

- [Some useful info about deploying](https://www.mkdocs.org/user-guide/deploying-your-docs/)


thomasht86:
This look great! 
I added the dependencies to `docs` in `pyproject.toml`, so we can install with `uv sync --extra docs`. 

Some of the formatting of snippets need to be formatted. 
Examples:
Screenshot 2025-03-26 at 07 50 17

- Make top snippet inside "Example usage", like the others? 


Screenshot 2025-03-26 at 07 49 40

- Remove `>>>` and remove indentation

Screenshot 2025-03-26 at 07 51 44

- Remove $ so it can be copied more easily

Notebooks look great! 

Also, see if you can make the `builder` and `grouping` modules be right under `vespa.querybuilder` in the API-reference. 
Now, there are two extra levels in the hierarchy. 

Screenshot 2025-03-26 at 08 19 34

vadimkatsel:
've addressed all the formatting issues, but I'm currently stuck on resolving the extra levels in the API structure.

- The fastest solution I found is to remove the` autoapi` plugin and configure the navigation manually. However, this would require maintaining the API navigation manually and creating corresponding `.md` files with links. Any future changes to the API structure would then require additional updates, which I assume we want to avoid.

- The alternative is to modify the folder structure, but this isn't ideal either. If there are existing references to inner files, there's no guarantee that all links will be updated correctly after restructuring.

thomasht86:
Looks like you can configure [autoapi](https://mkdocs-autoapi.readthedocs.io/en/latest/usage/#including-and-ignoring-patterns) to ignore only that folder (and add the two modules explicitly).

thomasht86:
I pushed a commit to add the API-reference explicitly. These do not change very often, so it is no problem updating manually if/when we need to. 
One more feature request: With sphinx, we could run doctest, which autotested the docstrings in modules.
This is a good way of making sure there are no errors in them. 
It would be nice to have this with mkdocs as well. This [module](https://github.com/koaning/mktestdocs) seem to be able to help. 

Nit: 
Screenshot 2025-03-27 at 09 28 24

we should have an extra line between imports and code for all examples for good style.

vadimkatsel:
Okay, make sence, if they don't change so often. 

I'll work on the tests, when I'll arrive after lunch.

vadimkatsel:
image
It's all good with docstring tests, but there's something happening at md files.

vadimkatsel:
image
Now tests are green. : )

thomasht86:
Great! Don't worry - running gh actions is often like that for most of us :)

vespa/application.py
```diff
@@ -145,54 +152,30 @@ def asyncio(
...
-            with app.syncio() as sync_app:
-                response = sync_app.query(body=body)
-
-        See :class:`VespaSync` for more details.
 
-        :param connections: Number of allowed concurrent connections
-        :param total_timeout: Total timeout in secs.
-        :param compress (Union[str, bool], optional): Whether to compress the request body. Defaults to "auto", which will compress if the body is larger than 1024 bytes.
-        :return: Instance of Vespa asynchronous layer.
-        """
```
thomasht86, line 195: This should not be removed. 😬 
Causing notebook tests to fail - see https://github.com/vespa-engine/pyvespa/actions/runs/14124406342/job/39570315422

thomasht86 (changes requested):

vespa/application.py
vadimkatsel: Since I was manually working with replacements of docstrings I might’ve missed some lines.

vadimkatsel: I’ll compare it to master and check all the small mistakes I’ve might have done.

vadimkatsel:
image
So the gh action for deploying shows, that it should be built fine.

vadimkatsel:
It has warnings, but they doesn't seem to be critical.

PR #1048, opened 2025-04-07
Resolving feeding of new pages to the Vespa Search

vadimkatsel:
- Add new `safe_filename` hash function
- Use it for creating shorter names


vadimkatsel:
So after final edits here It is - updated feed split for the new pages of pyvespa docs.
@thomasht86 thanks for that much attention you were giving for answering my questions and thinking together!

thomasht86 (approved):
good job!

PR #1053, opened 2025-04-29
docs: Use correct function type for example

jesse-c:
Whilst this is for the async version of `feed_iterable`, thee callback isn't awaited on[^1].

[^1]: https://github.com/jesse-c/pyvespa/blob/master/vespa/application.py#L682


thomasht86:
You are absolutely right @jesse-c! 
Nice catch, and thank you for contributing 👍

jesse-c:
No worries! Thank you for accepting the change. 🙇🏻

Issue #1055, opened 2025-05-01
Async upserts not working with httpx==0.28.1


jlcsilva:
[Recent changes in httpx](https://github.com/encode/httpx/pull/3364) break upserts done with [`VespaAsync.update_data`](https://github.com/vespa-engine/pyvespa/blob/cd42087ac2759955661651c59abd63436ef8d109/vespa/application.py#L1994), despite the misleading 200 status code. 

As explainhed [here](https://github.com/encode/httpx/discussions/3559), if `params={}`, all query parameters are ignored. Thus, if there are no `kwargs={}`, the `create` query parameter is ignored.

My proposal is to pass the `create` parameter to the httpx client directly, rather than on the URL. The updated method would then be:

```
    async def update_data(
        self,
        schema: str,
        data_id: str,
        fields: Dict,
        create: bool = False,
        auto_assign: bool = True,
        namespace: str = None,
        groupname: str = None,
        semaphore: asyncio.Semaphore = None,
        **kwargs,
    ) -> VespaResponse:
        path = self.app.get_document_v1_path(
            id=data_id, schema=schema, namespace=namespace, group=groupname
        )
        end_point = "{}{}".format(
            self.app.end_point, path
        )
        if create:
            kwargs["create"] = str(create).lower()
        if auto_assign:
            vespa_format = {"fields": {k: {"assign": v} for k, v in fields.items()}}
        else:
            # Can not send 'id' in fields for partial update
            vespa_format = {"fields": {k: v for k, v in fields.items() if k != "id"}}
        if semaphore:
            async with semaphore:
                response = await self.httpx_client.put(
                    end_point, json=vespa_format, params=kwargs
                )
        else:
            response = await self.httpx_client.put(
                end_point, json=vespa_format, params=kwargs
            )
        return VespaResponse(
            json=response.json(),
            status_code=response.status_code,
            url=str(response.url),
            operation_type="update",
        )
```

thomasht86:
Closed in #1056

PR #1056, opened 2025-05-01
Fix async upserts for new httpx versions

jlcsilva:

Fixes issue #1055, by providing query parameters directly to httpx methods rather than via url string.


thomasht86:
Thanks for catching this so early @jlcsilva 🙌
Added an integration test, that fails without your change for `httpx==0.28.1` and passes without. 
I updated httpx manually to test, will update lockfile in separate PR later. 
Ignoring 429 on linkchecks for now.

thomasht86 (approved):

PR #1057, opened 2025-05-03
Guide notebook for query builder

YashasviMantha:
The current documentation for the pyvespa's query builder relies on [unit tests](https://github.com/vespa-engine/pyvespa/blob/master/tests/unit/test_qb.py) and a [notebook](https://github.com/vespa-engine/pyvespa/blob/master/docs/sphinx/source/query.ipynb), which, while comprehensive in their own ways, lack a holistic, application-based perspective.

It might be quite advantagous to have a dedicated guide on using the query builder that goes over a lot more functionality than the notebook and is based on the sample apps currently deployed.

In this context I wrote a guide on using the query builder that is based on the Cord19 sample app. The guide goes over a lot of the functionality already seen in the unit test cases.

Requesting for feedback on improvements on the guide after which it can potentially be merged.

Also want to understand the best way to showcase the `nearestNeighbor` and other embeddings based functionality (maybe through the search app?) considering the cord19 app lack embedders. Currently the guide needs to be enhanced on those.  

Tagging @thomasht86 as discussed.

### DIFF
--- a/docs/sphinx/source/query-builder.ipynb
+++ b/docs/sphinx/source/query-builder.ipynb
@@ -1,0 +1,489 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "WORK_IN_PROGRESS\n",
+    "\n",
+    "<picture>\n",
+    "  <source media=\"(prefers-color-scheme: dark)\" srcset=\"https://assets.vespa.ai/logos/Vespa-logo-green-RGB.svg\">\n",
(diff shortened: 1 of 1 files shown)


thomasht86:
Thank you @YashasviMantha! 
This looks great! 
I am sorry to inform you that we recently decided to shut down the cord-19 app, so it would be great if we could use the [vespa-documentation-search](https://github.com/vespa-cloud/vespa-documentation-search) application for the whole notebook. 
I can add another searchChain to that one that would allow for building more advanced queries to showcase the QueryBuilder.

YashasviMantha:
Oh thats too bad that cord 19 is shutting down. Happy to convert the notebook to the search app based.

Issue #1058, opened 2025-05-05
Support to add headers when using vespa operations


snurf198:
I found out that there is no way to add header but only query params when sending sending operation requests to vespa. But in my case, I want to use the basic authorization and add authorization header when sending the request. So I wanted to ask to consider adding the way to add headers to the existing api.

thomasht86:
Thank you!
This is a reasonable request, and not much work. 
Could also be a good first issue if you want to contribute @snurf198?
Lmk if you want to take it on and I can provide some guidance.

snurf198:
Yes, I would love to take it on! Providing me some guidance will definitely help.

thomasht86:
Great @snurf198 ! 

README should help you get set up with dev-environment (using astral.sh/uv is highly recommended).

You will need to add `additional_headers` as an optional argument (dict) to the `Vespa`-class, and make sure that that `Vespa.base_headers` is merged with the provided argument. Then it will be used by both `VespaSync` and `VespaAsync`-clients. Update docstring and example usage.
We also need tests for this: 
- Unit test: `tests/unit/test_application.py` - should test that headers are added. (Can use mock following existing pattern)
- Integration test: `tests/integration/test_integration_docker.py` - Add a new test class (if not any of the existing ones are suitable to reuse).

Good luck!

snurf198:
@thomasht86 I have added a draft pull request solving this issue. But I am having trouble while adding integration test. I don't know which integration test I need to add and how to add it. Can you help me for this?

thomasht86:
Thank you! Pushed a commit with an integration test now. 🙂

snurf198:
@thomasht86 Thanks! Now I opened a pull request

PR #1063, opened 2025-05-11
Add addtional headers to Vespa class

snurf198:
I added additional headers as an optional argument to the `Vespa` class and added a unittest related to it(#1058). 
However I don't know which integration test I need to add and how to do it. Can I get help for this one?


thomasht86 (approved):
Great! Thanks @snurf198 🙏

PR #1067, opened 2025-06-10
Add VespaMatchEvaluator

thomasht86:

This should make it easier to also evaluate only the match-phase more efficiently.


thomasht86:
Checks fail due to cord19 api, which is probably deprecated now. Will fix in another  PR.

glebashnik (approved):

PR #1088, opened 2025-07-03
Fix user auth input also from terminal

thomasht86:

Tried to think why this was implemented this way - but can't think of anything other than logging in interactively from notebook was the original use case. 

To test, run eg. 

```python
from vespa.deployment import VespaCloud
import os

vespa_cloud = VespaCloud(
    tenant="vespa-team",
    application="rag-blueprint",
    application_root=".",
)
```
from terminal (REPL or as script, should work).


kkraune (approved):

kkraune:
thanks!

PR #1090, opened 2025-07-07
Fix 'alias component: component_alias' feature

Plenitude-ai:
In the vespa documentation, it is [mentioned](https://docs.vespa.ai/en/reference/schema-reference.html#uri) that we can alias directly a component of a field, which is not yet possible in pyvespa.
I propose this very simple contribution, that would allow to use this feature in pyvespa, simply using the pattern "component: component_alias".
I added some documentation for this.
In some places, this might break the spirit of some examples. Feel free to point this as over-documented and I'll remove it.

Automatic comment which I read and approved :

### DIFF
--- a/tests/unit/test_package.py
+++ b/tests/unit/test_package.py
@@ -1582,12 +1582,18 @@ class TestFieldAlias(unittest.TestCase):
                         alias=["single_aliased_field_alias"],
                     ),
                     Field(
+                        name="single_component_aliased_field",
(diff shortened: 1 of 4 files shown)


thomasht86 (approved):
Great! Thank you for your contribution! 👍

thomasht86:
Failing tests are unrelated (datasets version, will fix in subsequent PR)

Issue #946, opened 2024-10-02
Support for building query profiles


dainiusjocas:
Currently query profiles are only a key value mappings. That leaves a lot of functionality, e.g. query profile variants https://docs.vespa.ai/en/query-profiles.html#query-profile-variants, not supported.

Also, only one query profile is supported per application package. It would be great to support multiple.

PR #1091, opened 2025-07-21
Multiple query profiles

Plenitude-ai:
Vespa proposes to use different query profiles to ease the configuration on the client side, as well as versioning them and therefore be able to deploy new changes gradually, which I find very useful. It is not yet possible to have other query profiles aside from the 'default'.
This PR proposes BREAKING changes, for ApplicationPackage:
- `query_profile:QueryProfile` is now `query_profiles:Dict[str, QueryProfile]` (note the added s in the name of the attribute)
- `query_profile_type: QueryProfileType` is now `query_profile_types: Dict[str, QueryProfileType]`

I did not yet do the tests, but of course I commit to repair them if this PR is considered interesting value for the package :) 

Automatic message that I read and approve :

### DIFF
--- a/vespa/package.py
+++ b/vespa/package.py
@@ -3002,6 +3002,8 @@ class ApplicationPackage(object):
         schema: Optional[List[Schema]] = None,
         query_profile: Optional[QueryProfile] = None,
         query_profile_type: Optional[QueryProfileType] = None,
+        additional_query_profiles: Optional[Dict[str, QueryProfile]] = None,
+        additional_query_profile_types: Optional[Dict[str, QueryProfileType]] = None,
         stateless_model_evaluation: bool = False,
         create_schema_by_default: bool = True,
         create_query_profile_by_default: bool = True,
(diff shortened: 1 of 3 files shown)


thomasht86:
Thanks a lot! This looks like a good proposal, but we want to avoid breaking changes. I will take this as input and see if we can do it without breaking changes.

Plenitude-ai:
Yes I understand.
I tried to minimize the impact:
- the variables are not renamed
- it's still possible to pass a QueryProfile or QueryProfileType to the ApplicationPackage

This way, there should not be any breaking changes when building the App package
However, if one accesses the ApplicationPackage.query_profile, one will find a dict[str, QueryProfile] instead of a QueryProfile
In the same way, if one accesses the ApplicationPackage.query_profile_type, one will find a dict[str, QueryProfileType] instead of a QueryProfileType
I guess it's not that bad, but still breaking

If it's still an issue, maybe we could just create something like `additional_profiles`, `named_profiles`, `complementary_profiles` or `other_profiles`.
This could make sense, if we keep using the query_profile as root, and the other as additional query_profiles

Plenitude-ai:
I implemented the additional_query_profiles, so that it does not break anything
What do you think of it ?

thomasht86:
Hey!
With reference to this github issue: #946, your proposal would solve the last part. 
I would like us to solve the first part as well. 
It also seem a bit unintuitive to introduce the `additional_`- parameters. 
I have a WIP that could solve both, hope to get it in today.

Plenitude-ai:
Yes sure ! Thanks for your proposal this looks nice !

PR #1107, opened 2025-08-10
feat: add support for diversity

dainiusjocas:

Support for the [diversity](https://docs.vespa.ai/en/reference/schema-reference.html#diversity) configuration in the rank profiles.

### DIFF
--- a/vespa/package.py
+++ b/vespa/package.py
@@ -1431,6 +1431,41 @@ class Mutate(object):
         )
 
 
+class Diversity(object):
+    def __init__(self, attribute: str, min_groups: int) -> None:
+        r"""
+        Create a Vespa ranking diversity configuration.
+
+        This is an optional config that is used to guarantee diversity in the different query phases.
+        Check the `Vespa documentation
+        <https://docs.vespa.ai/en/reference/schema-reference.html#diversity>`__
+        for more detailed information about diversity configuration.
+
+        :param attribute: Which attribute to use when deciding diversity. The attribute must be a single-valued numeric, string or reference type.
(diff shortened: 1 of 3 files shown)


vespa/package.py
```diff
@@ -1431,6 +1431,41 @@ def __repr__(self) -> str:
...
+        Check the `Vespa documentation
+        <https://docs.vespa.ai/en/reference/schema-reference.html#diversity>`__
+        for more detailed information about diversity configuration.
+
+        :param attribute: Which attribute to use when deciding diversity. The attribute must be a single-valued numeric, string or reference type.
+        :param min_groups: Specifies the minimum number of groups returned from the phase.
+
+        >>> Diversity(attribute="popularity", min_groups=5)
+        Diversity('popularity', 5)
+        """
```
thomasht86, line 1449: Can we use google-style docstring like we do for the others, e.g

```python
        """
        Create a Vespa ranking diversity configuration.

        This is an optional config that is used to guarantee diversity in the different query phases.
        Check the [Vespa documentation](https://docs.vespa.ai/en/reference/schema-reference.html#diversity)
        for more detailed information about diversity configuration.

        Args:
            attribute (str): Which attribute to use when deciding diversity. The attribute must be a single-valued numeric, string or reference type.
            min_groups (int): Specifies the minimum number of groups returned from the phase.

        Returns:
            Diversity: A diversity configuration instance.

        Example:
            ```
            Diversity(attribute="popularity", min_groups=5)
            Diversity('popularity', 5)
            ```
        """
```

dainiusjocas: Done. Somehow missed the comments style 🤦

thomasht86 (approved):

PR #1110, opened 2025-08-12
Support perf environment

### LINKED ISSUE #1108: allow to deploy to perf clusters
Currently it doesn't seem to be possible to deploy a perf vespa cluster using pyvespa.
This would be good to have here too, not only in the cli

thomasht86:

Closes #1108

### DIFF
--- a/vespa/deployment.py
+++ b/vespa/deployment.py
@@ -1046,6 +1046,7 @@ class VespaCloud(VespaDeployment):
     ) -> None:
         """
         Delete the specified instance from the development environment in the Vespa Cloud.
+
         To delete a production instance, you must submit a new deployment with `deployment-removal` added to the 'validation-overrides.xml'.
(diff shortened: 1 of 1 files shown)


vespa/deployment.py
```diff
@@ -1011,11 +1035,15 @@ def deploy_from_disk(
...
         )
         return app
 
-    def delete(self, instance: Optional[str] = "default") -> None:
+    def delete(
+        self, instance: Optional[str] = "default", environment: Optional[str] = "dev"
+    ) -> None:
         """
         Delete the specified instance from the development environment in the Vespa Cloud.
         To delete a production instance, you must submit a new deployment with `deployment-removal` added to the 'validation-overrides.xml'.
```
sebastiannberg, line 1049: think it could be a good idea to mention dev or perf in the docstring here like we do in the other docstrings as well

sebastiannberg (approved):

PR #1114, opened 2025-08-29
Thomasht86/pyi stubs for lsp

thomasht86:

Ideally, we should autogenerate this in a GHA, but as long as we run the script when updating tags, it will do for now. 
and of course, ideally we would have explicit (valid) args for each tag, but haven't found a way to do that automatically yet.


glebashnik:
@thomasht86 Does it make sense to mention autogeneration in readme?

thomasht86:
It would, but I think we can do it in Github Action quite soon, so OK if we do that instead?

PR #1115, opened 2025-09-04
add integration test with key_location arg.

thomasht86:

FYI @pkobziak - not able to reproduce your issue from yesterday. 
Suspect that it has due to formatting of key file (lineendings or similar).


sebastiannberg (approved):
tests run, looks good!

PR #1121, opened 2025-09-17
fix: add the missing export_to_onnx method

dainiusjocas:

The support for adding ONNX models to application package is broken.

So, either the `export_to_onnx` method needs to be implemented, or the dance with temporary files needs to be reworked.

This PR add a missing method.

On the good note: it is possible to add ONNX model to the schema.

version 0.61.0 

@thomasht86

### DIFF
--- a/vespa/package.py
+++ b/vespa/package.py
@@ -1791,6 +1791,9 @@ class OnnxModel(object):
             repr(self.outputs),
         )
 
+    def export_to_onnx(self, output_path: Path) -> None:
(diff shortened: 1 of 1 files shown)


dainiusjocas:
Yes, stateless model evaluation. 

I wanted to prepare a notebook that could help when debugging models used in ranking.

There is yet another such missing method on parameter serialization, but I'm sure it is going to be obvious once a test is added for a stateless model evaluation.

PR #1128, opened 2025-09-30
Fixed hardcoded id_field in VespaMatchEvaluator

lrjball:

In VespaMatchEvaluator, the recall param is currently hard-coded to use `'id'` instead of the `id_field` param, so doesn't work with custom id fields. This minor change fixes that.


lrjball:
Actually never mind, looks like this line is getting deleted by #1125 anyway

thomasht86:
Thanks anyway - this should be fixed in #1125 as you could see. We also needed to make the `id_field`-argument mandatory there, which should make it less prone to misunderstandings.

PR #1130, opened 2025-10-01
Added deprecated warning when trying to use `type` in `Summary`

sebastiannberg:

Also added unit tests for this, in a new class `TestSummary`


thomasht86 (approved):
👍

PR #1136, opened 2025-10-07
Make it easier to reuse connections

### LINKED ISSUE #967: Make it easier to reuse connections
Currently in our fasthtml demo app, each call to Vespa uses the pattern

```
    async def get_something_from_vespa(self, doc_id: str) -> str:
        async with self.app.asyncio(connections=1) as session:
            ...
```

This ends up opening a new connection, doing TLS handshake for every request.

Quoting the doc, the session "should be used as a session manager" but this means we can't really store and reuse it.

Perhaps we can pass a httpx client which can be (re)used for the session?

thomasht86:

Closes #967. Will make it easier to keep connections open to avoid TLS handshake overhead, when.

### DIFF
--- a/tests/integration/test_integration_vespa_cloud.py
+++ b/tests/integration/test_integration_vespa_cloud.py
@@ -421,6 +421,36 @@ class TestConnectionReuse(unittest.TestCase):
         # Reusing should be faster
         self.assertLess(time_with_reuse, time_without_reuse)
 
+    def test_sync_connection_reuse_faster(self):
+        """Test that reusing a sync session is faster than creating new ones"""
+        import time
+
+        # Without reuse - create new session each time
+        start = time.time()
+        for _ in range(5):
+            with self.app.syncio() as session:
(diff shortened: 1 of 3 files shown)


andreer:
I _think_ I agree with copilot, at least if I understand it correctly we now make it possible to screw up by letting you say
... client: myClient, owns_session: True) - so make this a hidden field initialized based on whether "session" was passed and not a param?
Also is it possible to ensure that if the client param is passed, we fail if it is None? or impossible to separate those cases

andreer (approved):

thomasht86:
> I _think_ I agree with copilot, at least if I understand it correctly we now make it possible to screw up by letting you say ... client: myClient, owns_session: True) - so make this a hidden field initialized based on whether "session" was passed and not a param? Also is it possible to ensure that if the client param is passed, we fail if it is None? or impossible to separate those cases

Yes, me too. Removed in last 2 commits.

andreer (approved):

PR #1156, opened 2025-10-28
Allow document type names to differ from schema names in multi-document types schemas

PeriLara:

When creating an application with multiple document types within a single schema, the current implementation automatically uses the schema name as the document type name, which creates naming conflicts and limits flexibility.
This PR:
- Modifies schema-to-SD conversion logic
- Updates service XML generation in the case of multi document types in one schema
- Changes schema / document instantiation

However, this solution requires significant changes to core components. I wanted to:

Verify if this approach aligns with Vespa's architecture
Get feedback on whether there might be a simpler solution
Confirm if this is functionality that would be accepted upstream

### DIFF
--- a/vespa/package.py
+++ b/vespa/package.py
@@ -1918,6 +1918,9 @@ class Schema(object):
 
         self.stemming = kwargs.get("stemming", None)
 
+    def get_sd_name(self):
+        return self.document.name or self.name
+
     def add_fields(self, *fields: Field) -> None:
         """
         Add `Field` to the Schema's `Document`.
@@ -1999,7 +2002,7 @@ class Schema(object):
         schema_template = env.get_template("schema.txt")
         return schema_template.render(
             schema_name=self.name,
-            document_name=self.document.name or self.name,
+            document_name=self.get_sd_name(),
             schema=self,
             document=self.document,
             fieldsets=self.fieldsets,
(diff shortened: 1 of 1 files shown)


thomasht86:
Thanks a lot. 
This introduces a breaking change by making the `name`-parameter to `Document`-class mandatory. 
I think we should make it optional, and fall back to the name of the schema if not provided. 
That should make the tests pass. 

You can run unit tests with:

```bash
uv sync --extra dev 
uv run pytest tests/unit
```

thomasht86 (changes requested):
See above comment. 
Let me know if you need more input.

vespa/package.py
```diff
@@ -3421,7 +3442,7 @@ def to_zip(self) -> BytesIO:
 
             for schema in self.schemas:
                 zip_archive.writestr(
-                    "schemas/{}.sd".format(schema.name),
+                    "schemas/{}.sd".format(schema.document.name),
```
thomasht86: We don't want to name the .sd-file after the document. Should revert this. 
It also makes integration tests fail.

thomasht86 (changes requested):
See failing integration test and comments. We should not rename .sd-file, but keep it to the schema name.

thomasht86:
You can also try to run any of these subsets of integration tests locally: 

```yaml
matrix:
        test-suite:
          [
            "test_integration_docker.py",
            "test_integration_grouping.py",
            "test_integration_queries.py",
            "test_integration_evaluation.py",
          ]
```
These do not need any credentials to run (in contrast to the ones running against cloud). 
But, they do take some time to run.

PR #1160, opened 2025-11-03
Add Classes for Nearest Neighbor Optimization

boeker:
This adds classes for analyzing and optimizing the nearest neighbor search: hitratio determination, recall computation, parameter suggestion.

@thomasht86 Looking for feedback in particular on the interfaces of the classes.

### DIFF
--- a/tests/unit/test_evaluator.py
+++ b/tests/unit/test_evaluator.py
@@ -7,6 +7,7 @@ from vespa.evaluation import (
     RandomHitsSamplingStrategy,
     VespaNNGlobalFilterHitratioEvaluator,
     VespaNNRecallEvaluator,
+    VespaQueryBenchmarker,
     VespaNNParameterOptimizer,
 )
 from dataclasses import dataclass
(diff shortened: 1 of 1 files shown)


vespa/evaluation.py
```diff
@@ -1698,3 +1698,1076 @@ def extract_features_from_hit(
...
+        app (Vespa): An instance of the Vespa application.
+        hits (int, optional): Number of targetHits determined by the ANN queries that will be supplied to the instance of this class.
+        buckets_per_percent (int, optional): How many buckets are created for every percent point, "resolution" of the suggestions. Defaults to 2.
+        print_progress (bool, optional): Whether to print progress information while determining suggestions. Defaults to False.
+    """
+
+    def __init__(
+        self,
+        app: Vespa,
+        hits: int,
```
thomasht86: can we rename this to `target_hits` to avoid confusion with regular `hits`-parameter?

thomasht86 (changes requested):

vespa/evaluation.py
boeker: I have updated the description to clarify that this used as the regular `hits` parameter for the recall computations and should be the same as the `targetHits` annotation used in the queries.

vespa/evaluation.py
```diff
@@ -1698,3 +1710,1236 @@ def extract_features_from_hit(
...
+            "suggestion": filter_first_exploration,
+            "benchmarks": benchmarks,
+            "recall_measurements": recall_measurements,
+        }
+
+        return report
+
+    def run(self) -> Dict[str, Any]:
+        # Distribute queries to buckets
+        self.determine_hit_ratios_and_distribute_to_buckets(self.queries)
```
thomasht86: Would be nice to add the buckets distribution to the output report from `run`. 
Then I think we would have all info in result report to create all the relevant plots, wouldn't we?

thomasht86: Seems like we would also need the `benchmark.x` for each parameter?

thomasht86 (approved):
great work! 👍

PR #1163, opened 2025-11-10
Allow to specify number of max concurrent queries in ANN tool

boeker:
The default argument `max_concurrent` of `query_many` is quite high (100) and overloads Vespa when run locally. Moreover, the function `execute_queries` uses this default value and does not allow to change it.

This allows one to specify the value in `execute_queries` (default 100), `VespaQueryBenchmarker` (default 10), and `VespaNNParameterOptimizer` (default 10).


thomasht86 (approved):
Nice!

PR #1164, opened 2025-11-10
Use certifi CA file in VespaAsync

boeker:
Unlike `VespaSync`, `VespaAsync` does not seem to use the certifi CA file by default. This means that, when adding a custom certificate to that file (for example the certificate used in system tests), `VespaSync` can be used while `VespaAsync` cannot. This fixes that.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,6 +24,7 @@ dependencies = [
     "python-dateutil",
     "fastcore>=1.7.8",
     "lxml",
+    "certifi",
 ]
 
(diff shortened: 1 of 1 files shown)


thomasht86 (changes requested):
We should add `certifi` (unconstrained) to `pyproject.toml` as well. 
It currently works, as other libraries have `certifi` as dependency, but nice to be explicit about it.

thomasht86 (approved):
👍

PR #1168, opened 2025-11-12
more friendly results output

thomasht86:

It was not easy to use the output of the benchmark and recall runs in the notebook as it was.
But they are important to interpret outputs. Added the values, bucket info and a method to get dict from the results.
Also merged them into one class, just splitting on the metric name. The class can also be reused for other purposes later (eg. weakand-tuning). 
Added unit tests for the new class.


boeker (approved):
👍

Issue #1170, opened 2025-11-18
`VespaNNParameterOptimizer` - save progress and resume if interrupted


thomasht86:
Running this might take many hours, and if your network connection is interrupted, it will fail, and you need to start all over again 😢 
Maybe we could accept a `run_name` or something similar, and save progress somewhere, and allow for resuming if existing `run_name` is passed?

PR #1173, opened 2025-11-20
Add checkpoint/resume functionality to `VespaNNParameterOptimizer`

thomasht86:

Closing #1170 

Key changes:

- Checkpoint mechanism saves optimization progress at each stage (bucket distribution, filterFirstExploration, filterFirstThreshold, approximateThreshold, postFilterThreshold)
- Buckets now store query indices instead of full query objects to reduce state file size
- New parameters `run_name` and `resume` control checkpoint naming and resume behavior
- Replaced print statements with logger calls throughout the optimizer
- Added validation of state against current parameters to avoid resumption of a run that were run with different input.

### DIFF
--- a/docs/sphinx/source/examples/ann-parameter-tuning-vespa-cloud.ipynb
+++ b/docs/sphinx/source/examples/ann-parameter-tuning-vespa-cloud.ipynb
@@ -28,7 +28,7 @@
     "\n",
     "Approximate Nearest Neighbor (ANN) search is a powerful way to make vector search scalable and efficient. In Vespa, this is implemented by building HNSW graphs for embedding fields.\n",
     "\n",
-    "There are many parameters that controls how an ANN query is processed in Vespa\n",
+    "There are many parameters that control how an ANN query is processed in Vespa\n",
     "\n",
     "## Why different strategies are needed\n",
     "\n",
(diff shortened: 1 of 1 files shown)


boeker:
Not sure if one maybe should just add the option to specify values for the parameters instead. 🤔  Then, if a value is specified, that stage is skipped and the given value is used. This achieves roughly the same and is much simpler since it is the benchmarks that take time while the initial hit-ratio determination is comparatively short.

If the tool takes so long that it becomes unbearable, one should probably make it work faster instead of working around the slowness. Since it should only be run once, it is ok to be slow I think, but it should not be unbearably slow. For example, one could limit the number of queries in a bucket. If there are more queries in a bucket than the limit specifies, one could just choose a random subset.

thomasht86:
> Don't we want to sample `max_queries_per_benchmark` queries _per bucket_? If we have 20.000 queries in a bucket, then we should sample 100 from these, but if we only have 90 in a bucket, then we should use all these 90 queries. Otherwise, we might get very few queries in a bucket benchmark, which means that that specific benchmark result will be reliable.
> 
> Ideally, we would want to get the same number of queries, say 100, in a bucket. This then allows to make a reasonable choice for the number of benchmark repetitions. For example 10 repetitions than would mean that we take 10 * 100 samples, which would be constant across all buckets.

I did consider that, but isn`t much easier for user to relate to total queries?
In the case you're suggesting the user can't control number of total queries (and thus runtime) directly, right? 

And don't we want the distribution of queries to resemble their real query load?
 
Considering that many queries may have same hit ratios, but look completely different and have different response times, it seems reasonable to try to match the distribution for more transferrable results? Though I do agree that this may cause very few queries for some buckets, but as long as they are warned and aware, isn't that OK?

boeker:
> 1. It is more important to have enough queries per bucket than to approximate total distribution.

Exactly. I don't think we would benefit from approximating the total distribution at this point. You could use this to get the overall average response time/recall, but we do not compute this anywhere.

> * More stable results for _all_ buckets is more important than more accurate results for the _few_ buckets with many queries.

Yes. Having a single inaccurate bucket could throw the results off (suggest a latency spike where there is none, for example). So one has to try to avoid that.

> Should we consider not testing all percentages?

We could do that. Here, one could either do this in general (leave out some buckets) or try this in a more targeted way. Here, one should try to do this with the `filterFirstExploration` suggestion first since this takes the most time and has the most optimization potential (Only the results for buckets with low hit ratios matter, so we only have to benchmark these.)

Edit: Another meaningful optimization could be to cache the exact results in recall computations.

PR #1175, opened 2025-12-04
Perf tests, token vs mlts peformance

sebastiannberg:


tests/perf/conftest.py
```diff
@@ -0,0 +1,92 @@
...
+            "mtls_url": mtls_url,
+            "token_url": token_url,
+            "token": secret_token,
+            "cert_path": cert_path,
+            "key_path": key_path,
+        }
+    finally:
+        token_app.delete_all_docs(
+            content_cluster_name="msmarco_content", schema=app_package.name
+        )
```
thomasht86: We can remove this as it takes some time and is redundant as app is deleted anyway

sebastiannberg:
nice!

PR #1177, opened 2025-12-09
Thomasht86/add model config

thomasht86:

This lays out the base for doing a lot simpler evaluation of a hybrid search Vespa application, using eg. `mteb` library, using the `get_hybrid_package` - method combined with one or several `ModelConfig` objects as main entry point. 
Think we also need a way to connect the query parameters/ functions in some way. 
Let's discuss tomorrow.

### DIFF
--- a/tests/unit/test_models.py
+++ b/tests/unit/test_models.py
@@ -526,7 +526,6 @@ class TestCreateSemanticRankProfile:
         similarity_func = profile.functions[0]
         assert "closeness(field, embedding)" in similarity_func.expression
         # Should have transformation for hamming distance
-        assert "1/(1 + " in similarity_func.expression
 
     def test_custom_profile_name(self):
         """Test semantic profile with custom name."""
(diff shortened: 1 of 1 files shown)


thomasht86:
I will fix unit tests - just committed a related change

vespa/models.py
```diff
@@ -0,0 +1,896 @@
...
+        # Validate embedding dimension
+        if self.embedding_dim <= 0:
+            raise ValueError(
+                f"embedding_dim must be positive, got {self.embedding_dim}"
+            )
+
+        # Override embedding_field_type to int8 if binarized is True
+        if self.binarized:
+            object.__setattr__(self, "embedding_field_type", "int8")
+            # Validate binarized embeddings require dimension divisible by 8
```
sebastiannberg: Would be nice to have some kind of logging or feedback when we override attributes to make it more clear what is going on and why, but not a requirement

sebastiannberg (approved):

PR #1193, opened 2026-01-09
Add short page with infos on ANN parameter tuning to front page

boeker:
This adds  a short page with infos on ANN parameter tuning to the front page. The goal is to make the ANN parameter tuning easier to find and then to forward the reader to the example.


thomasht86 (approved):
Nice!

PR #1237, opened 2026-02-23
Handle multiple searchable copies in postFilterThreshold suggestion

boeker:
The postFilterThreshold is adjusted to the number of searchable copies internally in Vespa. It is (roughly) divided by the number of searchable copies. This means that we have to multiply the suggestion by the number of searchable copies.

### DIFF
--- a/tests/unit/test_evaluator.py
+++ b/tests/unit/test_evaluator.py
@@ -3345,6 +3345,81 @@ class TestVespaNNRecallEvaluator(unittest.TestCase):
             delta=0.0001,
         )
 
+    def test_compute_recall_id_field(self):
+        response_exact = self.SuccessfullMockVespaResponse(
+            [
+                {"id": "1", "fields": {"id": "1"}},
+                {"id": "2", "fields": {"id": "2"}},
+                {"id": "3", "fields": {"id": "3"}},
+                {"id": "4", "fields": {"id": "4"}},
+                {"id": "5", "fields": {"id": "5"}},
+            ]
+        )
+        self.assertAlmostEqual(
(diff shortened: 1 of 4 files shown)


vespa/evaluation/_base.py
```diff
@@ -1847,6 +1858,16 @@ def extract_from_trace(obj: dict, type_name: str):
 
         return all_hit_ratios
 
+    """
```
thomasht86: Should put docstring inside method

vespa/evaluation/_base.py
```diff
@@ -1847,6 +1858,16 @@ def extract_from_trace(obj: dict, type_name: str):
...
+    Returns number of searchable copies determined during hit-ratio computation.
+
+    Returns:
+        int: Number of searchable copies used by Vespa application.
+    """
+
+    def get_searchable_copies(self):
```
thomasht86: does it always return an int, or can it return None? 
Consider adding return type hint

tests/unit/test_evaluator.py
```diff
@@ -3221,7 +3221,28 @@ class MockVespaApp:
...
+                self, hits=[], _total_count=None, _timing=None, _status_code=200
+            ):
+                super().__init__(hits, _total_count, _timing, _status_code)
+
+            def get_json(self):
+                super_trace = super().get_json()
+                super_trace["trace"]["children"][0]["[0]"]["global_filter"][
+                    "upper_limit"
+                ] = 0.489
+                super_trace["trace"]["children"][0]["[0]"]["global_filter"][
```
thomasht86: duplicated..?

Issue #1184, opened 2025-12-18
Support for custom java bundles


dainiusjocas:
Right now there is no good way to add custom java components at all.

I'd expect that they could work similarly like custom ML modes:
- file paths in the application package
- on package construction they are copied into the final zip file

PR #1278, opened 2026-05-06
thomasht86/include files java bundles tests

thomasht86:

In #1270 we added support for adding files to app package. 
One of the more common use cases for this is adding Java bundles, ref #1184. 
Adding a test and example for doing this with the new feature from #1270.

I used Claude for this, but tested and validated thoroughly. 
Only adding test and docs, so high confidence.


BrageHK (approved):
Looks good. The ipynb files seems automatically formatted, which is unnecessary, but it does not make any difference.

PR #1295, opened 2026-05-29
feat: test and better err msg

BrageHK:

The test works on Vespa Cloud, but the tearDown function does not actually work. I just copied the one from the existing test.

No AI was used for writing this code.


vespa/deployment.py
```diff
@@ -1053,6 +1053,9 @@ def wait_for_prod_deployment(
                 skip_reason = status.get("skipReason")
                 if skip_reason:
                     print(f"Build skipped: {skip_reason}", file=self.output)
+                absorbed_build = status.get("absorbedIntoBuild")
+                if absorbed_build:
+                    print(f"Build batched, check logs for build {absorbed_build} for more info.")
```
sebastiannberg: add `file=self.output` here in case someone reroutes the output of the class to another file or place than stdout

thomasht86 (approved):
Tested manually 👍

PR #1316, opened 2026-06-25
Thomasht86/auto-dependency-updates

thomasht86:

**Why**

The uv.lock file isn't used by our users (they resolve from the loose bounds in pyproject.toml), but it's worth committing for reproducible, faster CI.

Dependabot opens lockfile-based version PRs several times a week. Each needs manual approval and usually a full CI run to confirm nothing breaks. That's a recurring hassle and a lot of redundant, slow, costly CI runs.

**What**

Replace the stream of Dependabot version PRs with a single scheduled job:

- update-deps.yml — weekly (Mondays) uv lock --upgrade to the latest versions allowed by pyproject.toml, gated on unit tests, opening one PR on deps/weekly-uv-upgrade. The uv.lock change triggers the full CI matrix; merge once green. No manual oversight unless something fails.
- integration-cloud.yml / notebooks-cloud.yml — added uv.lock to their pull_request path filters so the cloud suites also run on the deps PR.
- dependabot.yml — drops the pip ecosystem (now redundant); keeps GitHub Actions updates, grouped into one weekly PR. CVE coverage stays via Dependabot security updates in repo settings.
- TEST_STRATEGY.md — documents the testing + dependency strategy and the lockfile rationale.

Have added necessary secret and variable to repo.

FYI @odosk

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,16 @@
+# Python version bumps are handled by update-deps.yml (weekly uv re-lock), so no
+# pip ecosystem here. Security alerts stay on via repo settings, independent of
+# this file. Actions are kept current here, grouped into one PR.
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
+    commit-message:
+      prefix: "chore(ci)"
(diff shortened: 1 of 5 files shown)


.github/dependabot.yml
esolitos, line 1: 🤔  This is not necessary, right? 
Renovate should already handle this.

esolitos (approved):

thomasht86:
Converting to draft - need some more alignment first, and also taking into consideration renovate-settings at org level.

PR #1329, opened 2026-07-29
Fix worker limiting in feed_async_iterable

oystein-dev:
The semaphore which was supposed to limit the number of concurrent connections was only held during task creation. This has been changed so that the semaphore is instead held while the task is running. The test for `feed_async_iterable` has also been updated to check that the semaphore is passed to `feed_data_point`/`update_data`/`delete_data`.

I have done some manual benchmarks to see if the change limits the number of concurrent connections, but I don't know if this should be added as a unit test.

### DIFF
--- a/tests/unit/test_application.py
+++ b/tests/unit/test_application.py
@@ -2,6 +2,7 @@
 
 import json
 import unittest
+import asyncio
 
 import pytest
 from unittest.mock import PropertyMock, patch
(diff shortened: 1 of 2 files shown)


thomasht86:
Great! Would be nice if you could add the tests that this limits concurrent connections 🙏 
CI fail is transient, I think

thomasht86:
Great - test failure is unrelated (changed error type in httpr). Will fix in separate PR

thomasht86 (approved):
Great!

PR #1331, opened 2026-07-30
Fix error responses for data operations in VespaAsync

oystein-dev:
Error responses from the vespa api might not be json. This would previously give a json decode error. Now the functions instead return {"message": response.text} with the correct error code.


vespa/application.py
thomasht86, line 1994: we should use it here as well

thomasht86 (approved):
Nice!

PR #1336, opened 2026-08-04
Correct validation for tenant, app-name and instance as cloud

thomasht86:
> [!WARNING]
> This code was written by Claude Opus 5, but has been tested and verified by me (@thomasht86).

Correct validation for tenant, app-name and instance as cloud.

The application package name was validated with an ad-hoc rule (`name[0].isalpha() and name.islower() and len(name) <= 20 and name.isalnum()`), and the Vespa Cloud tenant, application and instance names were not validated at all.

The rules were retrieved and verified against `controller-server/src/main/java/com/yahoo/vespa/hosted/controller/api/identifiers/` in the `cloud` repo.

## Vespa Cloud names

New module `vespa/validation.py` implements the controller rules, applied in `VespaCloud` on construction and on `deploy` / `deploy_to_prod` / `deploy_from_disk`:

- Must start with a lowercase letter, may only contain lowercase letters, digits and dashes, no double-dashes, may not end with a dash, 1-40 characters.
- `api` is reserved for all three; `default` for tenant and application only; instance names ending in `-t` are reserved for testers.
- `validate_application_name(..., legacy_endpoints=True)` applies the stricter 20 character limit.

Two details differ from the issue description, both confirmed in the source:

| | pattern | `api` | `default` | other |
|---|---|---|---|---|
| `TenantId` | 40 | rejected | rejected | — |
| `ApplicationId` | 40, or 20 if `legacyEndpointsSupported` | rejected | rejected | — |
| `InstanceId` | 40 (no legacy variant) | rejected | **allowed** | `-t` suffix reserved |

1. **`default` is valid for instances.** `InstanceId extends SerializedIdentifier`, not `NonDefaultIdentifier` (only `TenantId` and `ApplicationId` do), and `default` is the standard instance name in Vespa Cloud.
2. **The legacy 20-char limit is application-only.** `TenantId.validate` and `InstanceId.validate` always use `strictPattern`; only `ApplicationId.validate` takes the `legacyEndpointsSupported` flag. `ApplicationController.withNewInstance` separately rejects instance names ending in `-t`.

The character rules are written as `[a-z][a-z0-9]*(-[a-z0-9]+)*` rather than transcribing the controller's `[a-z](-?[a-z0-9]+)*`. Same language, but the original's nested quantifiers backtrack exponentially in Python on a failing 40-character input. A unit test asserts both patterns accept and reject exactly the same names.

## Application package names

The cloud rules deliberately **do not** apply to `ApplicationPackage(name=...)`, which is not a Vespa Cloud application name — the tenant/application/instance names are passed to `VespaCloud` separately and validated there. Applying them would needlessly force self-hosted users onto cloud naming by rejecting the underscores that are idiomatic in Vespa schema names.

Package names are instead validated against what pyvespa actually derives from them (`[a-z][a-z0-9_-]*`, max 40), which allows underscores and the cloud-reserved `api` / `default`:

- The **default schema name**, a Vespa identifier `[a-zA-Z_][a-zA-Z0-9_]*` per `SchemaParser.jj`, once dashes are replaced by underscores — hence the required leading letter.
- The **`<name>_container` / `<name>_content` cluster ids**, which are `xsd:NCName` in services.xml and end up verbatim in Vespa Cloud endpoint names, where `Endpoint.sanitize` only translates `_` to `-` and does not lowercase — hence lowercase only and the length limit.
- The **Docker container name** used by `VespaDocker`.

Uppercase is not allowed, even though Vespa itself permits it in schema names and NCName cluster ids, because of that non-lowercasing `Endpoint.sanitize`. A test asserts the closure property: everything the package validator accepts yields a valid schema name and valid cluster ids.

| | validated against | underscores | `api` / `default` |
|---|---|---|---|
| `ApplicationPackage(name=...)` | `[a-z][a-z0-9_-]*`, ≤40 | yes | allowed |
| `VespaCloud(tenant=, application=, instance=)` | cloud rules | no | reserved |
| `VespaDocker` | nothing beyond the package name | yes | allowed |

## Behaviour changes

- `ApplicationPackage(name="my-app")` and `name="my_app"` are now accepted (both were rejected before). A dashed package name gets a default schema named with underscores (`my-app` -> schema `my_app`), since schema names cannot contain dashes.
- Application package names may now be up to 40 characters, was 20.
- A pre-existing Vespa Cloud tenant or application whose name contains underscores now raises in `VespaCloud`. The controller enforces `strictPattern` only at creation time, so such names can exist — but deploying to them would have failed anyway.
- `tests/unit/test_deployment.py` fixtures renamed from `test_tenant` / `test_app` to `test-tenant` / `test-app`, which the new rules require.

## Testing

- New `tests/unit/test_validation.py`, plus extended `TestValidAppName`.
- 709 unit + mktestdocs tests pass; `ruff check` clean.
- Integration tests not run (need Docker/Cloud credentials), but every tenant/application/instance literal in `tests/integration/` was checked against the new rules and is valid.

### DIFF
--- a/docs/sphinx/source/api/summary.md
+++ b/docs/sphinx/source/api/summary.md
@@ -10,3 +10,4 @@
             * [builder](vespa/querybuilder/builder/builder.md)
         * [grouping](vespa/querybuilder/grouping/index.md)
             * [grouping](vespa/querybuilder/grouping/grouping.md)
+    * [validation](vespa/validation.md)
--- a/docs/sphinx/source/api/vespa/validation.md
+++ b/docs/sphinx/source/api/vespa/validation.md
@@ -1,0 +1,1 @@
+::: vespa.validation
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -76,6 +76,7 @@ nav:
           - Builder: api/vespa/querybuilder/builder/builder.md
           - Grouping: api/vespa/querybuilder/grouping/grouping.md
       - Throttling: api/vespa/throttling.md
+      - Validation: api/vespa/validation.md
 plugins:
     - search
     - autorefs
@@ -155,3 +156,4 @@ plugins:
                 - api/vespa/querybuilder/builder/builder.md: QueryBuilder DSL
                 - api/vespa/querybuilder/grouping/grouping.md: Grouping queries
                 - api/vespa/throttling.md: Adaptive feed throttling
+                - api/vespa/validation.md: Validation of Vespa Cloud tenant, application and instance names
--- a/tests/unit/test_deployment.py
+++ b/tests/unit/test_deployment.py
@@ -3,22 +3,14 @@ from tempfile import TemporaryDirectory
 import os
 from unittest.mock import patch, MagicMock
 
-from urllib3.exceptions import HTTPError
-
-from vespa.deployment import VespaCloud, DataplaneToken
-from vespa.package import (
-    ApplicationPackage,
-    AuthClient,
-    ContainerCluster,
-    Nodes,
-    Parameter,
-)
+from vespa.deployment import VespaCloud
+from vespa.package import ApplicationPackage, AuthClient, Parameter
 
 
 class TestVespaCloud(unittest.TestCase):
     def setUp(self):
-        self.tenant = "test_tenant"
-        self.application = "test_app"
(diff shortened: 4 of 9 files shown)


BrageHK (approved):
Great work!

PR #1337, opened 2026-08-09
chore: accept httpr client in asyncio context

kriogenia:
Extremely simple change to fix the typings of the `asyncio` context builder to accept a `httpr` client too. That is what the underline `VespaAsync` init accepts so it is perfectly compatible without additional changes.


thomasht86 (approved):
Thank you!

PR #1340, opened 2026-08-24
Add pyvespa code-path lane to the performance tests

sebastiannberg:
## Why

The existing k6 test measures Vespa endpoints over raw HTTP, so it never exercises pyvespa's own code paths — a regression in the client (httpr changes, serialization, retry/compression layers) would go unnoticed. We also want to compare what pyvespa delivers against the raw-HTTP ceiling k6 represents.

## What

- Second lane driving the real pyvespa code paths against the same token/mTLS endpoints: `VespaSync.feed_data_point` (thread closed loop), `VespaAsync.feed_data_point` (asyncio closed loop), `feed_iterable`, and `feed_async_iterable`.
- Both lanes reduce to one shared `LaneResult` schema and assertion set (`utils/metrics.py`); load shape lives in `utils/workloads.py` and is injected into the k6 script via env vars so the lanes stay comparable.
- k6 runs at 200 VUs as the ceiling; each pyvespa method runs at an educated-guess near-optimal concurrency. pyvespa thresholds are deliberately loose until CI baselines exist.
- `k6_summary_to_prom.py` generalized into `reports_to_prom.py`: every (lane, method, transport) combination exports as one labeled Prometheus series for future Grafana dashboards.

🤖 AI declaration: _[Claude Code](https://claude.com/claude-code) wrote this PR title and description. The code is partly written with Claude; I have reviewed all code and commits myself._


thomasht86 (approved):
Great to have both lanes tested! 👍

PR #1341, opened 2026-08-24
chore: fix optional type hints

kriogenia:
Little change to method signatures were an optional is valid, and the default is `None`.  Also removal of an unused variable.

---


thomasht86 (approved):
Thanks! Noticed a few other methods need typing as well. I can follow up with those. We probably want to include type check in CI at one point.

PR #1342, opened 2026-08-24
feat: accept retry policies in async query

kriogenia:
As we discussed in Friday's catch-up @thomasht86 .

Here's the little PR to support configurable retries in `AsyncVespa.query` using [tenacity's retrying blocks](https://tenacity.readthedocs.io/en/latest/#retrying-code-block). The same approach could be followed for the rest of methods with the `@retry` decorator (and for the sync request with retry), but for now I kept the PR contained only on this one for faster delivery and discussion. I could help to complete the migration to this style for the rest of methods in the future.

This new version works exactly the same as the old one, with the additional overriding capabilities. Same default retry and all.

---

### DIFF
--- a/tests/unit/test_application.py
+++ b/tests/unit/test_application.py
@@ -25,6 +25,7 @@ from tenacity import (
     RetryError,
     wait_none,
     stop_after_attempt,
+    retry_if_result,
 )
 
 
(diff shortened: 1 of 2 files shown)


thomasht86:
Thank you - this is great. 
There is one path that might fail though; if users specify a retry callback. 

Adding a test such as the one below:

```python
async def test_query_honours_retry_error_callback(self):
    """A policy with retry_error_callback should return the callback's value."""
    app = Vespa(url="http://localhost", port=8080)
    vespa_async = VespaAsync(app)
    vespa_async._make_request = AsyncMock(side_effect=RuntimeError("fail"))

    policy = AsyncRetrying(
        wait=wait_none(),
        stop=stop_after_attempt(2),
        retry_error_callback=lambda state: "fallback",
    )

    result = await vespa_async.query(body={"yql": "x"}, retry=policy)

    assert result == "fallback"
    assert vespa_async._make_request.await_count == 2
```
fails with 

`UnboundLocalError: cannot access local variable 'query_response' where it is not associated with a value`

Could be solved by wrapping the query inside an async method, called through the retry. 

```python
async def _do_query() -> VespaQueryResponse:
    response = await self._make_request(
        "POST",
        self.app.search_end_point,
        json_data=body,
        params=kwargs,
        headers={"Accept": "application/cbor"},
    )
    return VespaQueryResponse(
        json=_response_json(response),
        status_code=response.status_code,
        url=str(response.url),
    )
return await retry(_do_query)
 ```
 
 I also think we should avoid naming the variable `retry` to avoid shadowing the retry-decorator. 
 Let me know if you follow me on this, and are willing to make the changes.

**Review by thomasht86 — CHANGES_REQUESTED** [review]
see comment above

**Comment by kriogenia** [comment]
@thomasht86 
Great suggestion and I think that invoking the AsyncRetrying object directly is better, closer to the decorator behavior. Update the PR with your suggestion, adding also the test to monitor that behavior and also the response building.

File: tests/unit/test_application.py [hunk 1]
```diff
@@ -1184,6 +1191,90 @@ async def test_update_data_async_falls_back_when_json_parsing_fails(self):
...
+        )
+
+        r = await vespa_async.query(
+            body={"yql": "select * from sources * where true"},
+            retry_policy=AsyncRetrying(wait=wait_none(), stop=stop_after_attempt(3)),
+        )
+
+        assert isinstance(r, VespaQueryResponse)
+        assert r.status_code == 200
+        vespa_async._make_request.await_count == 2
```
thomasht86:
```suggestion
        assert vespa_async._make_request.await_count == 2
```

thomasht86 (changes requested):
Great - one minor fix to add assertion in test, and we can merge this!

thomasht86 (approved):

Issue #1344, opened 2026-08-27
Unify and refactor retry strategies to module level


thomasht86:
## Background

#1342 made `VespaAsync.query` accept a caller-supplied `retry_policy: AsyncRetrying`, replacing its `@retry` decorator with direct invocation. The author noted the same treatment could be extended to the other retrying methods, and offered to help.

Before doing that method-by-method, it's worth defining the retry strategies once in a module-level place. Surveying the current state, the policies have drifted apart in ways that are hard to see when each one is spelled out inline at its call site.

## Current state

| Site | Strategy |
|---|---|
| `VespaSync._request_with_retry` (`application.py:1507`) — backs all sync ops incl. `Vespa.query` | Hand-rolled loop, not tenacity. `num_retries_429 = 10`, wait `0.1 * 1.618**attempt + uniform(0, 1)`. Retries 429 + connection errors |
| `VespaAsync.query` (`application.py:2288`) | `wait_random_exponential(multiplier=1.5, max=60)`, `stop_after_attempt(5)`. Exceptions only — **not** 429/503. Overridable via `retry_policy` |
| `VespaAsync.feed_data_point` (`application.py:2342`) | Stacked pair. Outer: any-exception + 503, `wait_exponential(1)`, `stop_after_attempt(3)`, `retry_error_callback=callback_docv1`. Inner: 429, `wait_random_exponential(multiplier=1, max=3)`, **no `stop`** |
| `VespaAsync.delete_data` / `get_data` / `update_data` | Same outer layer. Inner: 429, `wait_exponential(multiplier=1, max=10)`, **no `stop`** |
| `Vespa.visit` → `visit_request` (`application.py:1924`) | `retry_if_exception_type(HTTPError)`, `stop_after_attempt(3)` |
| `VespaCloud.get_connection_response_with_retry` (`deployment.py:1593`) | `stop_after_attempt(3)`, `wait_exponential(multiplier=1, max=3)` |
| `models.py:271` `_validate_single_url` | `stop_after_attempt(3)`, `wait_exponential(multiplier=1, min=1, max=10)` |

## Concrete problems this surfaces

**1. Unbounded retry on sustained 429.** The inner 429 decorator on all four document methods omits `stop`, so it inherits tenacity's default `stop_never`. Verified by introspecting the wrapped policies:

```
feed_data_point:
   layer0: stop=stop_after_attempt  wait=wait_exponential
   layer1: stop=_stop_never         wait=wait_random_exponential
[...]

PR #1346, opened 2026-09-03
refactor: define all retry strategies once in vespa.retries (#1344)

thomasht86:
## Context

> 🤖  Code written by Claude, reviewed and tested by me

Retry strategies have drifted across the codebase, and we want to gather them in one place, both to make them easier to change, and to reduce introducing new strategies and/or strategy drift in calling methods.

## What's changed? 

Every tenacity policy in the library now lives in vespa/retries.py as a module-level Retrying/AsyncRetrying constant. Call sites use POLICY.wraps or POLICY.copy()(fn, ...), so the constants are never mutated and tests have a stable patch target.

Behaviour-preserving, with two deliberate exceptions called out below.

- QUERY_RETRY: default for VespaAsync.query (5 attempts, random exp wait).
- DOCV1_RETRY + THROTTLE_RETRY: the two-layer policy for the async document/v1 methods, now applied once in a shared _docv1_request helper instead of four copies of stacked decorators.
- SYNC_REQUEST_RETRY: replaces the hand-rolled loop in VespaSync._request_with_retry; wait_golden_jitter reproduces the old 0.1 * 1.618**n + uniform(0, 1) formula and num_retries_429 still bounds it.
- VISIT_RETRY, CONTROL_PLANE_RETRY, URL_VALIDATION_RETRY: the decorators from Vespa.visit, VespaCloud and vespa.models, unchanged.

Deliberate changes:
- THROTTLE_RETRY spells out stop=stop_never. The 429 retry was already unbounded by omission; it is now explicit and documented, because it is the only backpressure mechanism on feed_async_iterable.
- The 429 wait differed between feed_data_point (wait_random_exponential(multiplier=1, max=3)) and its three siblings (wait_exponential(multiplier=1, max=10)). All four now share wait_random_exponential(multiplier=1, max=10): jittered so concurrent tasks that hit 429 together do not retry in lockstep, with the larger 10s cap. Release note: 429 retries on the async document API now use random exponential backoff capped at 10s.

Also, minor fixes:
- test_query_exhausts_retries[None-5] patched a tenacity private that had no effect and slept for real (9-34s). It now patches vespa.application.QUERY_RETRY and runs in ~0.1s.
- New tests/unit/test_retries.py covers policy shape, sync 429/connection retry, unbounded docv1 429, 503 exhaustion and exception re-raise.
- query.ipynb no longer claims queries retry 429/500/503/504 three times, which was true of neither client. API docs gain a vespa.retries page.

### DIFF
--- a/vespa/retries.py
+++ b/vespa/retries.py
@@ -125,63 +125,70 @@ RETRY_ON_503_OR_EXCEPTION = retry_any(
 
 # --------------------------------------------------------------------------- #
 # Policies
+#
+# Each constant is followed by a string literal so that both Sphinx autodoc and
+# mkdocstrings (griffe) pick it up as the attribute's docstring.
 # --------------------------------------------------------------------------- #
 
-#: Default for ``VespaAsync.query``: retry any exception up to 5 attempts with
-#: random exponential backoff. Does not inspect the status code.
 QUERY_RETRY = AsyncRetrying(
     wait=wait_random_exponential(multiplier=1.5, max=60),
     stop=stop_after_attempt(5),
 )
+"""Default for ``VespaAsync.query``: retry any exception up to 5 attempts with
+random exponential backoff. Does not inspect the status code."""
 
-#: Outer layer for async document/v1 operations (feed/get/update/delete): retry
-#: any exception or a 503 response up to 3 attempts. On exhaustion the last
-#: exception is re-raised or the last response is returned, never ``RetryError``.
 DOCV1_RETRY = AsyncRetrying(
     wait=wait_exponential(multiplier=1),
     retry=RETRY_ON_503_OR_EXCEPTION,
     stop=stop_after_attempt(3),
     retry_error_callback=_return_last_outcome,
 )
+"""Outer layer for async document/v1 operations (feed/get/update/delete): retry
+any exception or a 503 response up to 3 attempts. On exhaustion the last
+exception is re-raised or the last response is returned, never ``RetryError``."""
 
-#: Inner layer for async document/v1 operations: retry a 429 response.
-#: Deliberately unbounded (``stop_never``) so that sustained backpressure from
-#: Vespa slows the feed down instead of failing it. ``feed_async_iterable`` has
-#: no adaptive throttler, so this is the only backpressure mechanism on that path.
 THROTTLE_RETRY = AsyncRetrying(
-    wait=wait_exponential(multiplier=1, max=10),
+    wait=wait_random_exponential(multiplier=1, max=10),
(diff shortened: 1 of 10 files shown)


vespa/retries.py
```diff
@@ -0,0 +1,187 @@
...
+
+RETRY_ON_429 = retry_if_status(429)
+RETRY_ON_503_OR_EXCEPTION = retry_any(
+    retry_if_exception(_retry_on_any_exception), retry_if_status(503)
+)
+
+
+# --------------------------------------------------------------------------- #
+# Policies
+# --------------------------------------------------------------------------- #
```
sebastiannberg, line 131: The policy constants don't render on the built html page. docs build uses mkdocs + griffe, which only picks up a string literal on the line after the assignment as an attribute docstring. Undocumented members are then hidden by mkdocstrings' default `show_if_no_docstring: false`.

Repro:
```bash
uv run mkdocs build --site-dir docs/sphinx/build
# then inspect retries.html
```
Only `is_connection_error`, `retry_if_status`, `wait_golden_jitter` appear.

Fix: convert the `#:` comments to string literals under each constant (keeps `RETRY_ON_429` and friends hidden), or set `show_if_no_docstring: true` in `retries.md`.

vespa/retries.py
```diff
@@ -0,0 +1,187 @@
...
+
+#: Inner layer for async document/v1 operations: retry a 429 response.
+#: Deliberately unbounded (``stop_never``) so that sustained backpressure from
+#: Vespa slows the feed down instead of failing it. ``feed_async_iterable`` has
+#: no adaptive throttler, so this is the only backpressure mechanism on that path.
+THROTTLE_RETRY = AsyncRetrying(
+    wait=wait_exponential(multiplier=1, max=10),
+    retry=RETRY_ON_429,
+    stop=stop_never,
+)
```
sebastiannberg, line 154: Consider: `wait_random_exponential(multiplier=1, max=10)` here. `feed_data_point` had jitter before, and it is the one called in bulk. Without jitter, every task that gets a 429 at the same time sleeps the same 1, 2, 4, 8, 10... seconds and retries in lockstep, forever with `stop_never`. The test would then check the wait type instead of `multiplier`.

sebastiannberg (dismissed):

vespa/retries.py
thomasht86: good point!

vespa/retries.py
thomasht86: thanks!

sebastiannberg (approved):

PR #1348, opened 2026-09-17
feat: support Vespa Cloud production tests

thomasht86:
> Code written by Claude (Fable 5.1). Understood, tested and verified by me

This PR adds support for defining production tests with pyvespa following Vespa CLI additions in https://github.com/vespa-engine/vespa/pull/37761 and https://github.com/vespa-engine/vespa/pull/37799

Example: 

```python
from vespa.package import (
    ApplicationPackage,
    DeploymentConfiguration, Region, Delay, Test,
    ProductionTest,
)

app_package = ApplicationPackage(
    name="myapp",
    deployment_config=DeploymentConfiguration(
        environment="prod",
        steps=[
            Region("aws-us-east-1c"),
            Delay(minutes=10),
            Test("aws-us-east-1c"),
            Region("aws-eu-west-1a"),
        ],
    ),
    production_tests=[
        ProductionTest(
            name="cpu check",
            metric="cpu-utilization-container",
            duration="10m",
            max=85,
        ),
        ProductionTest(metric="query-latency-p95", duration="5m", max=200),
    ],
)
```

This will create a corresponding `deployment.xml` in correct location. 

- The changes are backwards compatible with existing API. 
- The `metric-presets.json` is vendored from vespacli python package. 
- Production tests are validated against this when submitted. 
- Decided _not_ to introduce dependency on `PyYaml`, so yaml files are not validated, but will be in Controller after submission. 
- Added a cloud integration test (but it is skipped because slow) that was run manually. Will send link to run in slack.
- A follow-up PR will add a scheduled workflow that opens a PR here when the upstream preset list changes.


BrageHK (approved):
Amazing!

PR #1350, opened 2026-09-20
Flag to store a deep copy of the request body in the response

dainiusjocas:

When replaying a lot of queries, it can be useful to be able to match the request with the response. 
When replaying queries in async mode, maintaining the mapping is pretty tricky and needs monkey patching pyvespa library.

`VespaResponse` object already has a slot for the `request_body`.

This PR adds an option parameter, to save a deep copy of the `request_body`.

The downside is that with big requests bodies it causes increased memory usage.


thomasht86 (approved):
Thanks for the contribution!
I understand the use case, and agree this is a useful parameter to add.

